PR #21, opened 2019-01-31
Java 11 blog post

aguibert:


lauracowen:
What you've written looks good but it doesn't tell the reader what the benefits of Java 11 are or why they should take the time to try it out (currently you just have instructions on how to try it out and then it all the things that might not work). It needs something more positive adding up-front.

Can you add into the opening intro (or insert a new sub-head after the opening intro) about why it's good to have Java 11? Rather than just diving straight into how to try it out. I've heard TLS and SPNEGO mentioned - can you do a first section on what the benefits of Java 11 are? Then provide the instructions on how to try it out (once you've convinced the reader that it's worth their time to bother). Then provide a quick list of caveats on what might not work. The JPMS section sounds like a benefit of Java 11 (even if it's not currently supported) so maybe that should be before the 'try it out' bit?

Some other quick comments:

- [ ] "OpenLiberty" should be two words (throughout).
- [ ] Typo in the second section: "applicaiton"
- [ ] Don't need to have a caption on the table - just make sure the text before the table clearly explains it and the table comes straight after that lead-in (it's not like in printed format where you have to direct people to find where the table is printed).

Can you talk with @NottyCode if you have any questions? He'll do the publishing whenever the support is ready. Thanks for writing this.

aguibert:
thanks for the comment Laura! 

I was still planning to add a "why should the reader care" bit highlighting the benefits, but I'm just going to briefly touch on these, as they are not necessarily Liberty specific and there is a lot of info already out there telling us what Java 11 (SE) offers.

aguibert:
NOTE: I still need to incorporate some of Laura's review comments, but here is an HTML build of what the blog post will look like. It doesn't look exactly like the final thing, but at least its better than the raw adoc.


[2019-01-21-java-11.html.txt](https://github.com/OpenLiberty/blogs/files/2822591/2019-01-21-java-11.html.txt)
(need to remove the .txt extension because github doesn't let you upload .html files apparently)

frowe:
Typo: The Java team threw **applicaiton** servers quite the curve-ball
Typo: **modue** com.foo {
Typo: **// Delcare** what packages are exported from this module
Typo: Oracle’s Official Java 11 **Migraiton** Guide

daliasheasha:
The content looks great! I had a few suggestions here and there:
- The Java team threw applicaiton servers -> The Java team threw application servers 
- has a decent lifespan -> has a production-friendly lifespan (or something else)
- lets look at the new stuff since Java 8 -> let's look at the new changes in Java 11 since Java 8
- There is tons of new stuff in the JDK since Java 8, but here is the stuff that I think-> There are tons of new features and changes in the JDK since Java 8, but here is the changes that I think
- You can new invoke HTTP requests -> You can now invoke HTTP requests
- Starting with `The Java Shell - JEP 222` and the following two titles, they are no longer underlined for some reason (missing formatting?) 
- The formatting for this sentence looks weird: `The java.lang.Process API has been enhanced so you can do some really convenient things without dipping into native code such as: - Getting detailed process info (pid, start cmd, cpu time, user) - List all child processes on the system - List child processes of the current process` I am assuming you'll be giving each item it's own line?

- Other Java VMs such as Hotspot are not supported due to some outstanding VM bugs. -> Other Java VMs, such as Hotspot, are not supported due to some outstanding VM bugs.
- Is there a reason you're not giving the https://openliberty.io/downloads/ page as an option to get openliberty? Also, I think the `How to set the Java used by OpenLiberty` section could use some numbering and sub-bullets so the steps are more highlighted. 

- Fyi, you have instructions on downloading openliberty and java 11 then you jump right into setting server config and running an application. Did you want to include a blurb about creating a new server? I don't know how much you assume the reader knows but for users using a different app server and not familiar with Liberty, that might be useful. 

- Before `There are a number of ways to control the Java executable used by the OpenLiberty server. The following list describes the order of precedence:`, you can add some context by saying `Next, you will need to set your OpenLiberty environment to use Java 11.`
- You can also add something like:
```
 In my environment, I went to my server's server.env config file `usr\servers\helloworld\server.env` and added the following:
 JAVA_HOME=C:\Java\jdk-11.0.2+9
```

- To confirm that OpenLiberty is using the intended Java level, you can check the first line of console output:-> To confirm that OpenLiberty is using the intended Java level, you can check the first line of console output when you start your server:
- To run an applicaiton on OpenLiberty that uses Java 11 -> To run an application on OpenLiberty that uses Java 11
- In Java 11 several modules -> In Java 11, several modules 

- I thought javax.activity was provided all the time not just ejbRemote-3.2?

- In Java 9 one of the biggest changes to Java was delivered, under JEP 261: The Module System.  -> In Java 9, one of the biggest changes to Java was delivered under JEP 261: The Module System.

aguibert:
thanks for the review comments @daliasheasha and @frowe! PR has been updated

drafts/2019-01-21-java-11.adoc
```diff
@@ -0,0 +1,293 @@
...
+Launching myServer (WebSphere Application Server 19.0.0.1/wlp-1.0.24.201901191232) on Eclipse OpenJ9 VM, version 11.0.2 (en_US)
+[AUDIT   ] CWWKE0001I: The server myServer has been launched.
+...
+----
+
+== Run a simple REST application on Java 11
+
+To run an application on Open Liberty that uses Java 11, we can start by defining the new compiler compliance level in the `build.gradle` file if using Gradle:
+
+[source,groovy]
```
kwsutter: I'd be consistent with the order of Gradle  and Maven examples.  Sometimes you lead with Maven, sometimes you lead with Gradle.

drafts/2019-01-21-java-11.adoc
```diff
@@ -0,0 +1,293 @@
...
+    <groupId>org.apache.maven.plugins</groupId>
+    <artifactId>maven-compiler-plugin</artifactId>
+    <version>3.8.0</version>
+    <configuration>
+        <release>11</release>
+    </configuration>
+</plugin>
+----
+
+NOTE: It is not required to set a compiler compliance level of Java 11 in order to run an existing application on Java 11, but in order to leverage any of the new language features you will need to do so.
```
kwsutter: Are you not going to explain the compiler compliance level and how to set that?  I think it would be good to explain both java 8 and 11 compliance using the jdk 11 compiler.

drafts/2019-01-21-java-11.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+To stop the server, run the command: `./gradlew libertyStop`
+
+== Java 11 changes that impact Open Liberty
+
+There have been a lot of good enhancements to Java between Java 8 and 11, such as local variables and the Flow API, but there are a few changes that are likely to cause problems for existing applications.
+
+=== Removal of Java EE and CORBA APIs from the JDK
+
+In Java 11, several modules (groupings of Java packages) were https://openjdk.java.net/jeps/320[removed from the JDK]. Your applications may have been relying on some of these removed packages, but the good news is that all of the removed packages are provided by existing Open Liberty features. Here is a table that shows what Open Liberty features to enable for each removed package:
```
kwsutter: Very nice!  Good information for the user.

drafts/2019-01-21-java-11.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+The JDK itself is now organized into Java modules, and it looks something like this:
+
+image::/img/blog/blog_java11_jpms.png[]
+
+The intent is for the entire Java ecosystem to eventually organize itself into Java modules, but for now Java modules are not defined in the JavaEE/JakartaEE/MicroProfile space. *While you can technically include a Java module in your Open Liberty application, the module-info descriptor will have no effect at the moment.*
+
+==== JPMS Access Restrictions
+
+While JPMS does introduce some interesting architectural capabilities and brings a long-needed modularity system into the JDK, it does so at the expense of several popular Java technologies, namely CDI and JPA. The reason for this is that Java modules cannot access members of non-exported packages from other Java modules, and Java code may no longer perform deep reflection (i.e. things that require `setAccessible(true)`) on members of Java modules. If either of these rules are violated, an `IllegalAccessException` is thrown.
```
kwsutter: This first paragraph may put some people on edge thinking that their JPA and CDI applications may no longer work...  The second paragraph alleviates those concerns, but taken out of context, it could cause some concern.

drafts/2019-01-21-java-11.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+Since this is such a major compatibility issue for most of the Java ecosystem, a "kill switch" (formally known as the `--illegal-access=permit` JVM option) has been introduced. It is enabled by default and allows code in unnamed modules to access and perform deep reflection on all named modules. It is important to note that eventually the kill switch will not be enabled by default, and later it will be removed entirely, so it is best to stop writing code that uses deep reflection and start migrating to alternative approaches.
```
kwsutter: Even this paragraph makes it sound like the user has control over whether this module access is allowed or not in the future...  But, even a careful developer could still hit problems if they are using a CDI or JPA implementation that still tries to go around these module restrictions.  Or, have our CDI and JPA implementations been all cleared up?  I'm not sure where we are at with running Open Liberty without the "kill switch"...

drafts/2019-01-21-java-11.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+== Useful references for Java 11 migration
+
+While this blog post focused mainly on Java 11 migration within the context of Liberty, there are many great resources out there for generic Java 11 migration, as well as Java 11 feature highlights. To name a few:
```
kwsutter: How about referencing our Migration Toolkit?  Dalia's team has many useful rules that can help detect issues with migrating to Java 11.

drafts/2019-01-21-java-11.adoc
aguibert: Last I checked, all of our tests still passed even with the kill switch turned off. This is mainly because the illegal access checks do not apply to unnamed modules (which all apps will be until we define JPMS in the EE/MP layer). Even if an application were to include a module-info today, at runtime the app would be running as an unnamed module because Liberty will not load the application as a module.

Eventually, if JPMS is defined in the application space, a developer can always open specific packages, or their entire module. This will allow current JPA/CDI implementations to deep reflect into the "open" code. As far as I'm aware, it will always be required for application developers to open up their code for injection and whatnot.

drafts/2019-01-21-java-11.adoc
aguibert: The migration toolkit is for WAS Liberty, and since this blog post is on openliberty.io, Alasdair and I were initially opposed to mentioning it. As I think about this again, I suppose its OK as long as its mixed in with other java 11 migration resources.

PR #22, opened 2019-02-02
Publish blog post on 4-weekly release cadence

NottyCode:


NottyCode:
I changed fixpack to updates. 
Although it was predictable before, the exact day couldn't be known, just that we would have one. So this is more predictable as you can predict the release dates for the rest of the year.
Added a sentence about zero migration.
I thought golden master was a fairly common term and wikipedia has a definition for it.

PR #26, opened 2019-02-08
Introduce code owners for stricter review process

kinueng:
We are close to delivering the capability to make a openliberty.io blog preview cards link to an external URLs.  I do not want people to use this capability without having a strict review process in place.

I have turned on requiring a review from code owners set by this PR.


kinueng:
Closing until we need this strict of a security setting

PR #72, opened 2019-08-01
edits to drafft blog post to be more prose

lauracowen:
How's this look, @brutif ? I was trying to make the intro bit less 'list-y' but I don't think I've done a great job so feel free to edit.

Would it make more sense to use the socialLogin plugin in the example code if that's easier to use?  

-- I figured I use oidc since that's that's what's new in OL. 

Can you add a summary sentence at the end of the whole post?

PR #73, opened 2019-08-02
Oidc blog 0802

brutif:


brutif:
Hi @lauracowen  I incorporated your changes from #72 and added other recommendations from Chunlong.

PR #79, opened 2019-08-15
Update README with instructions for blog tags

ellenwyllie:


lauracowen (commented):
I think I'd just say "To add tags..." (instead of "If you'd like to...") because we want people to keep the tags up-to-date. Otherwise, looks good to me. thx

lauracowen (approved):

PR #86, opened 2019-08-23
Update instructions for 3rd party blog posts

ellenwyllie:


lauracowen (changes requested):
How about formatting it as:

redirect_link: secure_url_to_3rd_party_post

(ie no quotes at all as that's what caused confusion?)

PR #87, opened 2019-08-27
Reactive streams 101

hutchig:
This pull request is a short article that introduces MicroProfile Reactive Streams Operators
and positions it in relation to java.util.streams and MicroProfile Reactive Messaging.

It is not really a 'here are lots of code examples to copy and get going' 
type of blog post as we don't ideally see MicroProfile Reactive Streams 
Operators as being an API that is targerted at the mainstream application 
programmer. So the aim is to position the API and discuss the thinking
behind why it exists and why it is designed the way it is.

### DIFF
--- a/drafts/2019-07-25-MicroProfileReactiveStreamsOperators.adoc
+++ b/drafts/2019-07-25-MicroProfileReactiveStreamsOperators.adoc
@@ -28,32 +28,32 @@ Gordon Hutchison https://twitter.com/gordhut
 
 == What is MicroProfile Reactive Streams Operators?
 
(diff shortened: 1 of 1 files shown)


hutchig:
I will review this Draft PR more and in my squad and then change to a 'real' PR.

hutchig:
Dear Laura, I think I would love your help in taking this text further, I have just re-read the comments and the article with a view to adding more but I think I need to defer to your writing skills/vision/requirements as there is a good chance I would be pushing the text about perhaps repeating topics but not necessary homing in on the target very well. I am extremely happy to spend any effort needed. So I will set it to a 'real' PR and pass on the baton - feel free to get me to loop again if you want to.

PR #127, opened 2019-09-13
added links and renamed to publish

lauracowen:


lauracowen:
Reviewed with @tomjenningss

PR #146, opened 2019-10-09
2019-10-11-configure-logs-JSON-format-190010.adoc

Yas-Franco:


lauracowen (changes requested):
- Can you change the title to something like "Configure server logs in JSON format to simplify dashboards on Open Liberty 19.0.0.10" (instead of "19.0.0.10 Blog release").
- Change the file name to `configure-logs-JSON-format-190010.adoc` to match.
- Match the `seo-title:` field to the same as the post title.
- Throughout, wherever there are code snippets can you put four dashes (`----`) on the line between `[source...]` and the code snippet, then another four dashes on the line after the code snippet (as in the 'Run your apps using 19.0.0.10' section).
- JSON section:
  - Where it says "message" and "log" in quotation marks, change the marks to backticks so they display as `message` and `log`. They're field names so display them literally as they would be seen.
  - Can you change the 'Logging and tracing' link URL to /docs/ref/general/#logging.html (instead of the page it currently targets which is the generated topic).
- Custom login modules
  - In the heading in parentheses, can you change to useing hte actual feature names (as shown in the first paragraph of the text below it) or remove the bit in parentheses completely (which I think is probably fine here anyway.
- Previews section
  - In the intro it says "not available in 19.0.0.8" but it should be "…19.0.0.10" (ie the current release).
  - Instead of just the MP feature name for the headings, can you think up more meaningful headings that capture the update then put the full name of the feature in parentheses instead of the abbreviated feature name? eg "Provide your own health check procedures (MicroProfile Health Check 2.1)". Then make sure the links to that section match the new headings.
  - Where there are method names (eg call() ) or actual feature names (eg mpHealth-2.1), can you put those in backticks so they display as 'code'. Also use backticks to highlight actual values of fields eg true, false.
  - use the [source, xml] (or [source, java]) style of markup for the code snippets in the MP Metrics section.

Sorry for being slow. Thank you!

drafts/configure-logs-JSON-format-190010.adoc
```diff
@@ -127,80 +132,37 @@ For more information:
...
-
-Depending on Jaeger's sampling settings `JAEGER_SAMPLER_TYPE` and `JAEGER_SAMPLER_PARAM`, Jaeger may not report every span created by the applications.
-
-For more information:
-
-* link:/https://github.com/jaegertracing/jaeger-client-java/blob/10c641f8df6316f1eac4d5b1715513275bcd724e/jaeger-core/README.md[Set up Jaeger settings using environment variables]
-
 [#hc21]
-== MicroProfile Health Check 2.1 (mpHealth-2.1)
+== Provide your own health check procedures (mpHealth-2.1)
```
lauracowen: The bit in brackets, can you spell out the full feature name? "(MicroProfile Health Check 2.1)"

drafts/configure-logs-JSON-format-190010.adoc
```diff
@@ -222,22 +184,24 @@ For more information:
 
 
 [#hm21]
-== MicroProfile Metrics 2.1 (mpMetrics-2.1)
+== Monitor microservice applications easily wth metrics (mpMetrics-2.1)
```
lauracowen: "(MicroProfile Metrics 2.1)"

drafts/configure-logs-JSON-format-190010.adoc
```diff
@@ -61,29 +62,32 @@ Or take a look at our link:/downloads/[Downloads page].
...
 An example of configuring JSON field name by adding the following to bootstrap.properties:
 [source]
+----
 com.ibm.ws.logging.json.field.mappings=trace:message:log,ibm_datetime:datetime
-
+----
 For more information on logging:
 
-* link:https://openliberty.io/docs/ref/config/#logging.html[Logging and tracing]
+* link:https://openliberty.io//docs/ref/general/#logging.html[Logging and tracing]
```
lauracowen: Think you've got an extra forward slash here but, with link formats (I should've spotted/mentioned this before, sorry), when you do a link to another page on the OpenLiberty.io website, can you use the format:
`link:/docs/ref/general/#logging.html` instead of the absolute URL (eg`link:https://openliberty.io/docs/ref/general/#logging.html`). Can you apply that to all the OL.io links in the post?

drafts/configure-logs-JSON-format-190010.adoc
```diff
@@ -61,29 +62,32 @@ Or take a look at our link:/downloads/[Downloads page].
 image::/img/blog/blog_btn_stack.svg[Ask a question on Stack Overflow, align="center"]
 
 [#jlog]
-== Configurable JSON log field names
+== Configurable JSON log field names (logging)
```
lauracowen: Don't think you need the "(logging)" bit of the heading - I don't think it adds meaning. The bits in parentheses are usually the name of the feature (not sure if logging is specific to a feature or just part of the OL kernel). Don't think it's a problem to be inconsistent and not have it for this heading.

lauracowen (changes requested):
Some minor changes but when they're done I think we're good. Thank you!

lauracowen (approved):
LGTM

PR #149, opened 2019-10-11
Add note about escaping characters in heading

lauracowen:
Steven, would you mind checking that this is technically correct before I publish, please? Do you know what other characters could need escaping in the heading? Thx


steven1046:
I think that's the only char we need to escape. We did have one other issue in the past where a blog had a `:` in the title so asciidoctor thought it was another metadata.

steven1046 (approved):

PR #180, opened 2019-11-03
first draft: prodcution experiance with open liberty

igbluz:
@lauracowen: please review the first draft for syntax and if you think it will meet the expectations


igbluz:
@lauracowen: looks good to me, you can do the final review and merge

PR #184, opened 2019-11-07
Create visual-studio-code-extension-for-microprofile-starter.adoc

Yas-Franco:


lauracowen (changes requested):
Sorry, these are going to sound picky. Some are syntax-related (functional), some are general style. Can you make the following changes, pls?

- [x] Remove quotation marks from around both the `title` and the `=` main heading of the post.
- [x] Change the capitalisation of the `title`, `seo-title`, and main heading so that "extension" is lower-case (not "Extension").
- [x] The paragraph that's in the body of the post won't get displayed. Can you move that whole text and use it to replace the values of the `seo-description` and `blog_description` fields in the metadata? The longer version is a better summary of the post.
- [x] Can you insert a space after `redirect_link:` (I have no idea if it's necessary but Jekyll builds can be a bit picky about the formatting of the frontmatter. Also makes it easier to read.)

yasmin-aumeeruddy:
It should be okay now

lauracowen (approved):

PR #189, opened 2019-11-08
Update blog to Metrics 2.2

Channyboy:
Metrics 2.1 is not being released.
Metrics 2.2 is the feature to be released.


Channyboy:
Please see PR #190 instead

PR #198, opened 2019-11-22
Igor

lauracowen:
Emily - would you mind giving this a quick review, pls? It's a post by Igor. I'd like to publish this afternoon if at all possible and you have time. Thx (ignore the json file).

Ellen - as sanity-check, pls could you review the json file? thx


lauracowen:
Published with #224 so closing this dupe.

PR #212, opened 2019-12-03
Fix asciidoc syntax in draft blog

steven1046:


lauracowen:
@steven1046 Is it necessary to have the quotation marks at all? In the latest version that I've just merged, Yasmin removed the quotation marks completely - will that fail to build/parse or is it fine as long as it's consistently with or without quotation marks on that parameter?

steven1046:
@lauracowen the description doesn't have to have quotes around it. But it does need a space after the title/github link.

lauracowen (approved):

lauracowen (approved):

PR #214, opened 2019-12-05
Update release-blog.adoc

Yas-Franco:


lauracowen (changes requested):
Can you remove the annotation caching section from the 'previews' part of the post, and then change the mention of annotation caching in the intro to mention GraphQL instead? Annotation caching is an internal that people shouldn't have to care about.

Otherwise, though, it's ready to send out to reviewers. Thanks!

lauracowen (approved):
LGTM

PR #219, opened 2019-12-06
revert commit that removed the reader role from post

lauracowen:
reader role is now GA so it can be mentioned in this post.


lauracowen:
Can you review this then ask @NottyCode to merge and rebuild, pls?

PR #223, opened 2019-12-13
Create prometheus-alertmanager.adoc

jennifer-c:
Prometheus Alertmanager blog


donbourne:
need to add more rules
change the default rule to be about heap size
call out that processCpuTime metric is in OL but not released yet
add a picture for what you see in slack when you get an alert

donbourne:
@jennifer-c , just a few comments:

- Configuring the Prometheus Alertmanager with Slack
    - step 2 - you say to edit alertmanager.yml to add your Slack webhook under global... should say what field name to use and possibly give an example
    
- Receiving Alerts via the Alertmanager
    - step 1 - looks like there is more than one "targets" in that file -- need to be more specific about which one
    
- Routes
    - should say somewhere which file the routes are configured in (alertmanager.yml?)
    
- Groups
    - in the blog you say "group_by: [ 'alerttype' ]", but in the sample alertmanager.yml you say "group_by: [ alerttype ]"

PR #248, opened 2020-01-22
Add api reference to social login blog

brutif:


brutif:
@lauracowen That's fine. I think it's going to read "API's documented [UserprofileManager Javadoc]",  but you might have intended:  "API's documented **in the** [UserprofileManager JavaDoc].

lauracowen:
> @lauracowen That's fine. I think it's going to read "API's documented [UserprofileManager Javadoc]", but you might have intended: "API's documented **in the** [UserprofileManager JavaDoc].

Good point. I've fixed that.

PR #277, opened 2020-02-12
Create 2020-02-19-CRUD-API_Using-Java-11

Yas-Franco:


lauracowen (changes requested):
- The URL looks wrong - I think it should be https://rieckpil.de/jakarta-ee-crud-api-tutorial/
- Can you change the URL slug to use only lower-case and only hyphens (`-`) not underscores (`_`)
- I think remove the "in a TDD manner" bit. Or spell out as "using test-driven development." but I'm not sure it's necessary.

yasmin-aumeeruddy:
The changes have now been made.

lauracowen (approved):
thank you!

PR #306, opened 2020-03-18
SameSite post style edit for review

lauracowen:
@pnicolucci Please can you review my style changes to ensure there's nothing that's changed the meaning? I've mostly focused on trying to ensure consistency in format and readability. In places, I've moved things round slightly to hopefully simplify, or added an extra sentence to clarify. But please make sure it's still correct. Thanks. When you're happy, we can schedule to publish it.

Also added the reason for using reference. Make sure you're happy with how I've phrased it.

### DIFF
--- a/drafts/samesite-cookies.adoc
+++ b/drafts/samesite-cookies.adoc
@@ -128,7 +128,7 @@ Any number of cookie names can contain the wildcard character at the end and, fo
 <samesite lax="*" strict="*"/>
 ----
 
-The `httpEndpoint` configuration also allows the use of a reference to the `samesite` configuration:
+If you have a SameSite configuration to use in multiple endpoints, you can create the configuration in one locaton and refer to it from the server configuration of each endpoint. For example:
 
 [source,xml]
 ----
--- a/drafts/samesite-cookies.adoc
+++ b/drafts/samesite-cookies.adoc
@@ -5,17 +5,17 @@ categories: blog
 author_picture: https://avatars1.githubusercontent.com/u/29307632
 author_github: https://github.com/pnicolucci
 seo-title: Setting the SameSite attribute on cookies with Open Liberty
-seo-description: You can now specify the `SameSite` cookie attribute for your application-defined cookies, as well as for the session and security cookies, in your server configuration and minimize application-level changes. The `SameSite` cookie attribute is used by web browsers to determine if a particular cookie should be sent with a request.
-blog_description: You can now specify the `SameSite` cookie attribute for your application-defined cookies, as well as for the session and security cookies, in your server configuration and minimize application-level changes. The `SameSite` cookie attribute is used by web browsers to determine if a particular cookie should be sent with a request.
+seo-description: SameSite attribute is used by web browsers to determine if a particular cookie should be sent with a request. You can now specify the `SameSite` cookie attribute for your application-defined cookies, session and security cookies, and any cookie added to the response by the server, in your server configuration and minimize application-level changes.
+blog_description: SameSite attribute is used by web browsers to determine if a particular cookie should be sent with a request. You can now specify the `SameSite` cookie attribute for your application-defined cookies, session and security cookies, and any cookie added to the response by the server, in your server configuration and minimize application-level changes.
 ---
 = Setting the SameSite attribute on cookies with Open Liberty
 Paul Nicolucci <https://github.com/pnicolucci>
 :imagesdir: /
 :url-prefix:
 
-On Open Liberty, you can now specify the `SameSite` cookie attribute for your application-defined cookies, as well as for the session and security cookies, in your server configuration and minimize application-level changes. The `SameSite` cookie attribute is used by web browsers to determine if a particular cookie should be sent with a request.
+On Open Liberty, you can now specify the `SameSite` cookie attribute for your application-defined cookies, as well as for the session and security cookies, and any cookie added to the response by the server, in your server configuration and minimize application-level changes. The `SameSite` cookie attribute is used by web browsers to determine if a particular cookie should be sent with a request.
 
-Until recently, if the attribute wasn't set, Google Chrome assumed the value of the attribute to be `None` and allowed third-party cookies to track users across multiple sites. A recent change to Chrome (and other web browsers will follow), however, has changed this default behavior so that cookies are only set for the domain specified in the cookie (`Lax`). To change this behaviour, the `SameSite` attribute must be explicitly set.
+Until recently, if the attribute wasn't set, Google Chrome assumed the value of the attribute to be `None` and allowed third-party cookies to track users across multiple sites. A recent change to Chrome (and other web browsers will follow), however, has changed this default behavior so that cookies are only sent for "same-site" and "cross-site" top-level navigations specified in the cookie (`Lax`). To change this behaviour, the `SameSite` attribute must be explicitly set.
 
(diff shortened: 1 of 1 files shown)


drafts/samesite-cookies.adoc
```diff
@@ -1,134 +1,157 @@
...
+In the following example, all cookies have the SameSite attribute set to `Lax`:
 
-Only one standalone wildcard character is allowed. The following is an invalid configuration:
+[source,xml]
+----
+<samesite lax="*"/>
+----
 
-   <samesite lax="*" strict="*"/>
+In the following example, sugar has the SameSite attribute set to `None` and the Secure attribute added, peanutButter has the SameSite attribute set to `Strict`, and all other cookies have the SameSite attribute set to `Lax`:
```
pnicolucci: can we add code formatting around `peanutButter` and `sugar` to be consistent ?

drafts/samesite-cookies.adoc
```diff
@@ -1,134 +1,157 @@
...
 
-The only available way to set the `SameSite` attribute using the Servlet specification is specifically defining a `Set-Cookie` header using the following two APIs:
+== Setting the SameSite attribute
 
-* HttpServletResponse
-. https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletResponse.html#setHeader-java.lang.String-java.lang.String-[HttpServletResponse.setHeader]
-. https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletResponse.html#addHeader-java.lang.String-java.lang.String-[HttpServletResponse.addHeader]
+Open Liberty is providing three new ways of setting the `SameSite` attribute because it is not possible to set the `SameSite` attribute using either the Servlet API or the link:{url-prefix}docs/ref/javaee/8/#package=javax/servlet/http/package-frame.html&class=javax/servlet/http/Cookie.html[Cookie API].
```
pnicolucci: I could not get any of the API links to work reviewing the document here, is that expected?

pnicolucci: Also the Cookie API is actually part of the Servlet APIs. We can set the samesite attibute using set/add header apis on the response object but can't do it directly on the Cookie API, cookie.setSameSite(String value) isn't available for instance, ideally it would be in the future then that cookie can just be added to the response object.

drafts/samesite-cookies.adoc
```diff
@@ -1,134 +1,157 @@
...
 
-If the `httpSession` and/or `webAppSecurity` configurations have not been set then given the above examples the Session cookie `SameSite` attribute value would be `Lax` as well as the Security cookies `SameSite` attribute value would be `Lax`.
+The link:{url-prefix}/docs/ref/config/#httpEndpoint.html#samesite[`httpEndpoint` configuration] also supports the wildcard character (`*`).
 
-Any number of cookie names can contain the wildcard character at the end and for any given cookie name the most specific pattern will be matched.
+In the following example, all cookies have the SameSite attribute set to `Lax`:
```
pnicolucci: I'd like to be a bit more clear here, all cookies is all cookies added by the application and the server, care should be taken if using a wildcard.

drafts/samesite-cookies.adoc
```diff
@@ -1,134 +1,157 @@
...
+On Open Liberty, you can now specify the `SameSite` cookie attribute for your application-defined cookies, as well as for the session and security cookies, in your server configuration and minimize application-level changes. The `SameSite` cookie attribute is used by web browsers to determine if a particular cookie should be sent with a request.
```
pnicolucci: Reading this again I'd like to word this a bit differently here and in the descriptions areas above. We allow the samesite attribute to be configured with wildcards and explicitly for not only application-defined,session,security cookies but for any cookie added to the response by the server as well.

drafts/samesite-cookies.adoc
```diff
@@ -1,134 +1,157 @@
...
 
-== What is the SameSite Cookie Attribute?
+Until recently, if the attribute wasn't set, Google Chrome assumed the value of the attribute to be `None` and allowed third-party cookies to track users across multiple sites. A recent change to Chrome (and other web browsers will follow), however, has changed this default behavior so that cookies are only set for the domain specified in the cookie (`Lax`). To change this behaviour, the `SameSite` attribute must be explicitly set.
```
pnicolucci: The domain and samesite attributes are actually used for two different things: domain determine which hosts to send the cookie to and samesite determine with which origin to send the cookie from. 

> ... however, has changed this default behavior so that cookies are only sent for "same-site" and "cross-site" top-level navigations ...

drafts/samesite-cookies.adoc
lauracowen: Good catch. Yes.

drafts/samesite-cookies.adoc
lauracowen: The links will resolve in the build. I'll test them all to make sure though.

I changed them to use our OL.io-hosted Javadoc.

lauracowen: > Also the Cookie API is actually part of the Servlet APIs. We can set the samesite attibute using set/add header apis on the response object but can't do it directly on the Cookie API, cookie.setSameSite(String value) isn't available for instance, ideally it would be in the future then that cookie can just be added to the response object.

So should that be re-phrased to something like ...can't use the Cookie API, which is part of the Servlet API...?

drafts/samesite-cookies.adoc
lauracowen: Will change "all cookies" to "all cookies added by the application and the server" - is that okay?

drafts/samesite-cookies.adoc
lauracowen: Extended the list to add ", and any cookie added to the response by the server" in that first para and in the metadata descriptions.

drafts/samesite-cookies.adoc
lauracowen: Sorry. :) Updated with your words.

drafts/samesite-cookies.adoc
pnicolucci: or we can just link to the javadoc for Cookie there?

drafts/samesite-cookies.adoc
```diff
@@ -1,134 +1,157 @@
...
-blog_description: A short article with simple configuration examples to demonstrate how to configure an Open Liberty server to set a SameSite attribute on cookies.
-additional_authors:
-- name: Paul Nicolucci
-  github: https://github.com/pnicolucci
-  image: https://avatars1.githubusercontent.com/u/29307632
+seo-title: Setting the SameSite attribute on cookies with Open Liberty
+seo-description: SameSite attribute is used by web browsers to determine if a particular cookie should be sent with a request. You can now specify the `SameSite` cookie attribute for your application-defined cookies, session and security cookies, and any cookie added to the response by the server, in your server configuration and minimize application-level changes.
+blog_description: SameSite attribute is used by web browsers to determine if a particular cookie should be sent with a request. You can now specify the `SameSite` cookie attribute for your application-defined cookies, session and security cookies, and any cookie added to the response by the server, in your server configuration and minimize application-level changes.
```
pnicolucci: should we add  ... and any cookie added to the response by the server to the blog_description to be consistent?

drafts/samesite-cookies.adoc
lauracowen: Removed "either the Servlet API or " - is that correct?

drafts/samesite-cookies.adoc
lauracowen: The two descriptions are identical, I think?

PR #339, opened 2020-04-23
Non-validating XML syntax and typo

lauracowen:
@NottyCode Does this updated syntax look right now?
@mrsaldana See my slack message but basically:

Using types as an attribute won’t validate against the server.xml schema; it should be enumerated nested elements. It matters because although it will run at runtime, it won’t validate in the tools (IDE or Admin Center). We shouldn't doc non-validating examples.


mrsaldana (approved):
Thanks for updating Laura, change looks good. Tested configuration and worked.

lauracowen:
thanks for testing it :)

PR #345, opened 2020-04-29
Asynchronous programming made resilient with MicroProfile Fault Tolerance

### LINKED ISSUE #373: Asynchronous programming blog post
Created for tracking blog post progress.

Joseph-Cass:
Regarding the following issue: https://github.com/OpenLiberty/open-liberty/issues/7753#event-3284300278


drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+== How you can use `@Asynchronous`
+So `@Asynchronous` sounds great, right? But how are we going to use it to make our asynchronous programs resilient? Let's go through 2 use-cases:
+
+1. Applying Fault Tolerance to an API which is already asynchronous
+2. Using `@Asynchronous` annotation to do two things in parallel with resilience
+
+=== 1. Applying to An Already Asynchronous API
+Another use is to apply Fault Tolerance to an API which is already asynchronous. We can do this with any API which does something Asynchronously and returns a `CompletionStage`.
+
+In this example, we'll use the JAX-RS client API for calling remote REST services, introduced in JAX-RS 2.1. We can build up a request to fetch a String from a given URL with a GET request and the return type is a `CompletionStage` of String:
```
Azquelt: These first two parts don't join together very well.

The intro says `@Asynchronous` makes your method run asynchronously. Then the blog goes straight into using `@Asynchronous` with an already asynchronous API so that fault tolerance is applied around the completion stage result.

There's nothing that links these two parts, so it feels like you're leading the reader along one path, then you jump over to something completely different.

I'm not sure of the best way to fix that. I think the problem is that if you talk about what `@Asynchronous` does, the bigger part is that it runs things asynchronously, and the user needs to be aware of that. However, in this blog, you wanted to lead with applying FT to an asynchronous API.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+
+But there's one in particular which is going to help you make your asynchronous methods resilient to faults: https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/Asynchronous.html[@Asynchronous].
+
+Let's show you what `@Asynchronous` does and how it does it.
+
+== What does `@Asynchronous` do?
+The `@Asynchronous` annotation makes your method run asynchronously, meaning that rather than running straightaway on the main thread when you call it, it will run some time later, usually on another thread. To do this, we need to add the `@Asynchronous` annotation to the method, and we change the method to return either a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage] (preferable) or a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html[Future]. In our method body, we can use `CompletableFuture.completedFuture()` to make our result compatible with the new return type.
```
Azquelt: I don't like saying "(preferable)" here without saying why.

`CompletionStage` isn't just a better `Future`, they do different things, so the "why" doesn't come until much later in the blog where we talk about how we can apply fault tolerance around a `CompletionStage` but we can't around a `Future`.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+   github: https://github.com/Azquelt
+   image: https://avatars3.githubusercontent.com/u/2011102?s=460&u=4754ac398d208d72a7aff0bdeab828562a192d53&v=4
+---
+= Asynchronous programming made resilient with MicroProfile Fault Tolerance
+Joseph Cass <https://github.com/Joseph-Cass>
+
+https://www.oracle.com/technetwork/java/javase/overview/java8-2100321.html[Java 8] is long-term support and https://www.jetbrains.com/lp/devecosystem-2019/java/[the most commonly used Java version]. It has many awesome features; one of which is https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage], which allows you to make your code https://www.oracle.com/technetwork/database/application-development/jdbc/learnmore/dev4798-5180524.pdf[asynchronous]. Great stuff! However, you might have the following questions:
+
+[.text-center]
+*_"How can I make my asynchronous methods resilient to faults? What if an asynchronous method unexpectedly returns an exception? What happens then?"_*
```
Azquelt: Take out "What happens then?"

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+blog_description: "MicroProfile Fault Tolerance @Asynchronous allows you to add additional functionality to @Timeout and @Bulkhead annotations and allows you to use the other Fault Tolerance annotations with other asynchronous APIs"
+additional_authors: 
+ - name: Andrew Rouse
+   github: https://github.com/Azquelt
+   image: https://avatars3.githubusercontent.com/u/2011102?s=460&u=4754ac398d208d72a7aff0bdeab828562a192d53&v=4
+---
+= Asynchronous programming made resilient with MicroProfile Fault Tolerance
+Joseph Cass <https://github.com/Joseph-Cass>
+
+https://www.oracle.com/technetwork/java/javase/overview/java8-2100321.html[Java 8] is long-term support and https://www.jetbrains.com/lp/devecosystem-2019/java/[the most commonly used Java version]. It has many awesome features; one of which is https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage], which allows you to make your code https://www.oracle.com/technetwork/database/application-development/jdbc/learnmore/dev4798-5180524.pdf[asynchronous]. Great stuff! However, you might have the following questions:
```
Azquelt: > It has many awesome features; one of which is CompletionStage, which allows you to make your code asynchronous.

I think this is a misleading statement. How about "`CompletionStage`, which provides new ways to write asynchronous code."

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: Yes, I agree this sounds better

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: Yes I agree. It currently implies that this wasn't possible before `CompletionStage`

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: I feel we could resolve this by rewording the last few paragraphs of the blog which explain the differences between the two (`CompletionStage` and `Future`) to sound more like a benefit of `CompletionStage` rather than a difference, then linking the word `preferable` to that part of the blog?

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Emily-Jiang: 'Another use' sounds out of place as it is the first use. You can update to" One usage is to apply ...".

Based on what @Azquelt said, I think if you rephrase the first one to focus on "Applying `@Asynchronous` to An Already Asynchronous method". By the way, it should not be Asynchronous API, it should be Asynchronous Method.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+
+Sometime later, your method runs and returns a `CompletionStage` or `Future`. Now, Fault Tolerance will go and complete the `CompletionStage` or `Future` object which was returned to the caller earlier, with the value that was returned from your method. You might be wondering *_"What if the returned result is an exception? What can I do?"_*.  Read along.
+
+== How you can use `@Asynchronous`
+So `@Asynchronous` sounds great, right? But how are we going to use it to make our asynchronous programs resilient? Let's go through 2 use-cases:
+
+1. Applying Fault Tolerance to an API which is already asynchronous
+2. Using `@Asynchronous` annotation to do two things in parallel with resilience
+
+=== 1. Applying to An Already Asynchronous API
```
Emily-Jiang: Asynchronous API -> Asynchronous Method

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+
+If we call the `clientDemo()` method, without any annotations, it works as we expect. We call it, receive a `CompletionStage<String>` (named `response` in the example), and then we can add an action to take when the `CompletionStage` completes.
+
+[source,java]
+----
+response.thenAccept(System.out::println);
+--> responseText
+----
+
+However, what happens if we want to use the `@Retry` annotation with our method? The answer is nothing.
```
Emily-Jiang: The answer is nothing. => The answer is that the method will not be retried. The explanation is as follows.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+	...
+
+}
+----
+
+That means that when we call this method and it returns a `CompletionStage`, it's only when that `CompletionStage` completes that Fault Tolerance will look at the result and decide whether to retry. So, if the request fails, the `CompletionStage` completes with an exception, Fault Tolerance decides that a Retry is needed and it calls the method again. As before, Fault Tolerance has intercepted the method call, so the `CompletionStage` returned to the caller is a different `CompletionStage` so that the caller doesn't get the result until all the Retries have been completed.
+
+So, to recap, to use Fault Tolerance with an asynchronous API you must:
+
+- **Return a `CompletionStage` from your method** - You can't do this with a `Future`, it must be a `CompletionStage`.
```
Emily-Jiang: The first part of this article does not mention anything about Future not applicable. You should explain why prior to this recap.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+
+		doSomethingWhichMightFail()
+		return CompletableFuture.completedFuture("serviceB");
+
+	}
+}
+----
+[#FT-interactions]
+== Interactions with other Fault Tolerance annotations
+We've covered running things in parallel and applying Fault Tolerance to asynchronous APIs, now let's look at the way using the `@Asynchronous` annotations impacts other Fault Tolerance annotations.
```
Emily-Jiang: I suggest to change aynchronous APIs to asynchronous methods.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+When you use the `@Asynchronous` and `@Bulkhead` annotations together, Fault Tolerance provides the option to queue up executions if the maximum number of executions are already running. This is allowed because any calling code has been written with the knowledge that the method is asynchronous and won't return immediately. 
+
+When you call the method, if there are less than the maximum concurrent executions running then your method is scheduled to run immediately, otherwise it's added to a queue. When one execution of the method finishes, if there are any on the queue then the first execution from the queue is started. If the queue itself is full, then the method fails with a `BulkheadException`.
+
+Just like the number of concurrent executions, the size of the queue can also be configured using the `waitingTaskQueue` parameter on the `@Bulkhead` annotation.
+
+== Deep dive: Asynchronous flow of execution
+This last section isn't critical for knowing how to use `@Asynchronous`, but it's useful stuff to know when working with MicroProfile Fault Tolerance. It covers how the flow of execution changes when using the `@Asynchronous` annotation compared to when it's not used.
```
Emily-Jiang: it's useful stuff -> it's useful

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,230 @@
...
+
+The next difference comes in the Bulkhead. As well as either accepting or rejecting the execution, the Bulkhead can also queue it to be run later. If the method is accepted by the Bulkhead, it is then scheduled to be run on another thread, rather than run immediately. 
+
+Another difference occurs with Timeout. When used with @Asynchronous, if the Timeout expires, the method is interrupted, and the execution skips forwards to the point highlighted in the diagram (From the "Timeout Expires" block). The result is then processed as if the method had finished with a TimeoutException.
+
+The last difference is that if there's a Fallback, it also runs asynchronously, so it's scheduled to run on another thread as well.
+
+If a method returns a `CompletionStage` rather than a `Future`, then there's one more difference compared to the synchronous flow of execution. If the method returns a value rather than throwing an exception, we wait until that `CompletionStage` completes before doing the rest of the Fault Tolerance logic. A few things to point out here:
```
Emily-Jiang: It is unclear what is the difference you tried to say. Do you mean if returning 'Future', we don't wait for Future to complete? It's not true though.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: Yes, I agree this sounds better

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: I agree, this is a little more clear.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: I'm happy to make this change, but think `method` should not be capitalised: https://www.oracle.com/java/technologies/glossary.html

What do you think?

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: Ah good spot Emily. I did not notice it still says "Another use"

I'll include a potential solution in the next draft (Draft 7) and we can see if it resolves this issue.

Azquelt: What was wrong with Asynchronous API?

"Applying Fault Tolerance to an Already Asynchronous Method" sounds a bit like you're applying fault tolerance to a method which is already annotated with `@Asynchronous`.

Whatever we say here wants to convey that you're calling some third party code which is going to do something asynchronously for you.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Azquelt: @Joseph-Cass If the heading is title cased (as it currently is), then "method" would be capitalised as a normal word.

However, it looks like the other headings in this article aren't title cased, I think this is just left over from when you were referring to it as AAAAA, so you should probably decapitalise everything for consistency.

Joseph-Cass: Ah yes, correct

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: "Applying FT to an already asynchronous method invocation"?

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+The last difference is that if there's a Fallback, it also runs asynchronously, so it's scheduled to run on another thread as well.
+
+[#CS-preference]
+=== Why is `CompletionStage` prefered?
+If a method returns a `CompletionStage` rather than a `Future`, then there's one more useful difference compared to the synchronous flow of execution. If the method returns a value rather than throwing an exception, the execution waits until that `CompletionStage` completes before doing the rest of the Fault Tolerance logic. There are a few benefits to point out here: 
+
+* The execution reserves a space on the Bulkhead and doesn't release it until after the returned `CompletionStage` completes. So, as far as the Bulkhead is concerned, it's still "executing" until the `CompletionStage` is complete. 
+* The timeout starts before we check if there’s space on the Bulkhead, at which point it might be queued. If it is queued, the time it spends queuing still counts towards the timeout. 
+* Similarly, the timeout isn't stopped until after the `CompletionStage` completes.
```
Azquelt: * The middle point is not unique to `CompletionStage`, the timeout is always started before an execution is queued.
* There's a missing point: if the method returns a `CompletionStage` that completes with an exception, Retry, CircuitBreaker and Fallback will act as if that exception was thrown from the method.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+The next difference comes in the Bulkhead. As well as either accepting or rejecting the execution, the Bulkhead can also queue it to be run later. If the method is accepted by the Bulkhead, it is then scheduled to be run on another thread, rather than run immediately. 
+
+Another difference occurs with Timeout. When used with @Asynchronous, if the Timeout expires, the method is interrupted, and the execution skips forwards to the point highlighted in the diagram (From the "Timeout Expires" block). The result is then processed as if the method had finished with a TimeoutException.
+
+The last difference is that if there's a Fallback, it also runs asynchronously, so it's scheduled to run on another thread as well.
+
+[#CS-preference]
+=== Why is `CompletionStage` prefered?
+If a method returns a `CompletionStage` rather than a `Future`, then there's one more useful difference compared to the synchronous flow of execution. If the method returns a value rather than throwing an exception, the execution waits until that `CompletionStage` completes before doing the rest of the Fault Tolerance logic. There are a few benefits to point out here:
```
Azquelt: I don't think this section really explains why `CompletionStage` is preferred. It just explains how the behaviour is different to `Future`, it doesn't make a good case as to why it's better.

I think there are two key advantages of `CompletionStage`:
1. Fault Tolerance policies are applied around the resulting completion stage rather than around the method call _which means you can apply fault tolerance around an existing API_. (I think the existing section does explain the different behaviour pretty well, but doesn't draw the link that these behaviours are necessary when calling an existing asynchronous API).
2. You can get the result from a CompletionStage without doing a blocking wait.

It might make sense to keep the detailed description of the differences here and name this section something more appropriate, but earlier in the article give a briefer explanation of the differences and draw that link between the CS behaviour and applying FT around an existing asynchronous API.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+But there's one in particular which is going to help you make your asynchronous methods resilient to faults: https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/Asynchronous.html[@Asynchronous].
+
+Let's show you what `@Asynchronous` does and how it does it.
+
+[#Asynchronous-logic]
+== What does `@Asynchronous` do?
+The `@Asynchronous` annotation makes your method run asynchronously, meaning that rather than running straightaway on the main thread when you call it, it will run some time later, usually on another thread. To do this, we need to add the `@Asynchronous` annotation to the method, and we change the method to return either a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage] (<<CS-preference, preferable>>) or a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html[Future]. In our method body, we can use `CompletableFuture.completedFuture()` to make our result compatible with the new return type.
```
Azquelt: I would change "To do this" to "To use it"

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,237 @@
...
+== How you can use `@Asynchronous`
+So `@Asynchronous` sounds great, right? But how can you use it to make your asynchronous programs resilient? Let's go through 2 use-cases:
+
+1. Applying Fault Tolerance to an API call which is already asynchronous
+2. Using the `@Asynchronous` annotation to do two things in parallel with resilience
+
+=== 1. Applying Fault Tolerance to an asynchronous API call
+One use is to apply Fault Tolerance to an asynchronous API call which returns a `CompletionStage`. Since `@Asynchronous` makes code run asynchronously <<Asynchronous-logic, as we've discussed>>, you may be wondering how this can help us with something which is already asynchronous? Well, because without it, you wouldn't be able to apply Fault Tolerance. Let's see why.
+
+Note: The reason why Fault Tolerance can only be applied to asynchronous method calls returning a `CompletionStage` and not a `Future` is https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.0.2/microprofile-fault-tolerance-spec.html#_interactions_when_returning_a_code_future_code[exaplined here].
```
Azquelt: You're linking to the spec which says how Future works, but I don't think it actually gives a reason as to why CompletionStage and Future are different.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,237 @@
...
+[source,java]
+----
+@GET
+@Path("/resource")
+public CompletionStage<String> fetchExample() { 
+
+	return CompletableFuture.completedFuture("responseText");
+
+}
+----
```
Azquelt: I don't think this is needed, the server-side code isn't relevant to the article.

If we do use it then
* "Fetching the following resource" should not be a header
* We should not have the method use the JAX-RS 2.1 asynchronous server-side API - I think that just adds confusion because it's asynchronous but not anything to do with fault tolerance.
* We should list it carefully as an _example_ of something which our client-side code _could_ be calling.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+If we call the `clientDemo()` method, without any annotations, it works as we expect. We call it, receive a `CompletionStage<String>` (named `response` in the example), and then we can add an action to take when the `CompletionStage` completes.
+
+[source,java]
+----
+response.thenAccept(System.out::println);
+--> responseText
+----
+
+However, what happens if we want to use the `@Retry` annotation with our method? The answer is that the method will not be retired.
```
Azquelt:
```suggestion
However, what happens if we want to use the `@Retry` annotation to retry any failed requests with our method? The answer is that the method will not be retried.
```

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Azquelt: Maybe change this section so that it's explaining the example rather than telling the user what they _have_ to do.

Azquelt:
I had a discussion with Joe and I think it would help resolve some of the comments raised if we change the first section to be clear that `@Asynchronous` does two things: runs the method asynchronously and applies FT around the result of the CompletionStage (if the method returns a `CompletionStage`). The application of FT around the CS is critical to the first use case, so I think we should list it as a first-class feature of the `@Asynchronous` annotation, rather than introducing it later.

I think this would help with some of the ordering issues in the article (we had a "recap" which was adding new information) and clearly delineate between "what `@Asynchronous` does" and "how to use it".

I'd also suggest changing the last section "Why is `CompletionStage` preferred" to instead be called "Limitations of using `@Asynchronous` and `Future`".

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+image::/img/blog/FT-basic-asynchronous-execution.png[Fault Tolerance asynchronous execution flow]
+
+Sometime later, your method runs and returns a `CompletionStage` or `Future`. Now, Fault Tolerance will go and complete the `CompletionStage` or `Future` object which was returned to the caller earlier, with the value that was returned from your method. You might be wondering *_"What if the returned result is an exception? What can I do?"_*.  Read along.
+
+== `@Asynchronous` Use cases 
+So `@Asynchronous` sounds great, right? But how can you use it to make your asynchronous programs resilient? Let's go through 2 use cases:
+
+1. <<Applying-to-asynch-api, Applying Fault Tolerance to an API call which is already asynchronous>>
+2. <<Running-methods-in-parallel, Using the `@Asynchronous` annotation to do two things in parallel with resilience>>
```
Emily-Jiang: Why only two things? I suggest to change `two` to `multiple`.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+- **Return a `CompletionStage` from your method** - You can't do this with a `Future`, it must be a `CompletionStage`.
+- **Use the `@Asynchronous` annotation** - Without it, the method will never throw an exception, even if it fails.
+
+When you do these two things, all the other Fault Tolerance logic is applied when the `CompletionStage` completes, rather than when the method returns.
+
+You can also use https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/microprofile-fault-tolerance-spec.html#_interactions_with_other_fault_tolerance_annotations[other Fault Tolerance annotations with '@Asynchronous'] to make your asynchronous method resilient. See the <<FT-interactions, Interactions with other Fault Tolerance annotation>> section below for more detail about how the annotations interact.
+
+[#Running-methods-in-parallel]
+=== 2. Let’s go parallel!
+To run two methods in parallel, you can write methods which call two other services, annotate them both with the `@Asynchronous` annotation, then call them like this:
```
Emily-Jiang: To run two methods -> To run multiple methods
call two other services -> call other services
annotate them both -> annotate them

Joseph-Cass: Yep, I like this idea :)

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: Lol, only just noticed the `retired`

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: Decided to remove the example

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+blog_description: "MicroProfile Fault Tolerance @Asynchronous allows you to add additional functionality to @Timeout and @Bulkhead annotations and allows you to use the other Fault Tolerance annotations with other asynchronous APIs"
+additional_authors: 
+ - name: Andrew Rouse
+   github: https://github.com/Azquelt
+   image: https://avatars3.githubusercontent.com/u/2011102?s=460&u=4754ac398d208d72a7aff0bdeab828562a192d53&v=4
+---
+= Asynchronous programming made resilient with MicroProfile Fault Tolerance
+Joseph Cass <https://github.com/Joseph-Cass>
+
+https://www.oracle.com/technetwork/java/javase/overview/java8-2100321.html[Java 8] is long-term support and https://www.jetbrains.com/lp/devecosystem-2019/java/[the most commonly used Java version]. It has many awesome features; one of which is https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage], which provides new ways to write https://www.oracle.com/technetwork/database/application-development/jdbc/learnmore/dev4798-5180524.pdf[asynchronous code]. Great stuff! However, you might have the following questions:
```
Charlotte-Holt: What about:

> Java 8 has long-term support and is the most commonly used Java version. It has many awesome features—one of which is the CompletionStage interface. CompletionStage provides new ways to write asynchronous code...

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+[.text-center]
```
Charlotte-Holt: I'm not sure I'd center this, but we can build it and see how it looks.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+*_"How can I make my asynchronous methods resilient to faults? What if an asynchronous method unexpectedly returns an exception?"_*
+
+https://github.com/eclipse/microprofile-fault-tolerance[MicroProfile Fault Tolerance] is here to help!
+
+There are many different annotations which MicroProfile Fault Tolerance provides for helping you deal with failure: https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/Retry.html[@Retry], https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/Timeout.html[@Timeout], https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/CircuitBreaker.html[@CircuitBreaker], https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/Bulkhead.html[@Bulkhead] and https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/Fallback.html[@Fallback].
```
Charlotte-Holt: >There are many different annotations which MicroProfile Fault Tolerance provides for helping you deal with failure:

To

>There are many different annotations that MicroProfile Fault Tolerance provides to help you deal with failure:

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+But there's one in particular which is going to help you make your asynchronous methods resilient to faults: https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/apidocs/org/eclipse/microprofile/faulttolerance/Asynchronous.html[@Asynchronous].
```
Charlotte-Holt: >But there's one in particular which is going to help you make your asynchronous methods resilient to faults: 

To

>But there's one, in particular, that's going to help you make your asynchronous methods resilient to faults:

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+Let's show you what `@Asynchronous` does and how it does it.
```
Charlotte-Holt: >Let's show you what `@Asynchronous` does and how it does it.

To

>Let's show you what `@Asynchronous` can do and how it does it.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+[#Asynchronous-logic]
+== What does `@Asynchronous` do?
+
+Adding the `@Asynchronous` annotation to a method does two things:
+
+1. Runs the method asynchronously
+2. Allows Fault Tolerance strategies to be applied to asynchronous methods
+
+Adding the `@Asynchronous` annotation to a method makes the method run asynchronously, meaning that rather than running straightaway on the main thread when it's called, it will run some time later, usually on another thread. To use it, you need to add the `@Asynchronous` annotation to a method, and must return either a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage] or a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html[Future]. You cannot apply Fault Tolerance to a returned `Future` (see why <<CS-preference, here>>), so `CompletionStage` is preferred.
```
Charlotte-Holt: >Adding the `@Asynchronous` annotation to a method makes the method run asynchronously, meaning that rather than running straightaway on the main thread when it's called, it will run some time later, usually on another thread.

To

>Adding the `@Asynchronous` annotation to a method makes the method run asynchronously, meaning that rather than the annotated method running straight away on the main thread when it's called, it will run sometime later, usually on another thread.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+----
+@Asynchronous
+public CompletionStage<String> serviceA() { 
+
+	return CompletableFuture.completedFuture("serviceA");
+
+}
+----
+
+Note: The example uses `CompletableFuture.completedFuture()` to make the result compatible with the `CompletionStage` return type (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html[CompletableFuture] is an implementation of the `CompletionStage` and `Future` Interfaces).
```
Charlotte-Holt: I might take out "Note", and just begin with "The previous example". I also think you might need a period after "return type"?

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+What happens when the `serviceA()` method is called?
+
+Well, usually when a method is called it returns a result. However, when the `@Asynchronous` annotation is used, Fault Tolerance gets in between. When the method is called, Fault Tolerance first schedules the method to run later on a different thread, then it returns a `CompletionStage` or `Future`. This isn't the `CompletionStage` or `Future` that's returned from the method, because the method hasn't run yet, it's an object created entirely by Fault Tolerance.
```
Charlotte-Holt: >Well, usually when a method is called it returns a result. However, when the `@Asynchronous` annotation is used, Fault Tolerance gets in between. When the method is called, Fault Tolerance first schedules the method to run later on a different thread, then it returns a `CompletionStage` or `Future`. This isn't the `CompletionStage` or `Future` that's returned from the method, because the method hasn't run yet, it's an object created entirely by Fault Tolerance. 

To

>Well, usually when a method is called it returns a result. However, when the `@Asynchronous` annotation is used, Fault Tolerance intervenes. When the annotated method is called, Fault Tolerance first schedules the method to run later on a different thread, then it returns a `CompletionStage` or `Future`. This isn't the `CompletionStage` or `Future` that's returned from the method, because the method hasn't run yet. It's an object that Fault Tolerance creates.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+What happens when the `serviceA()` method is called?
+
+Well, usually when a method is called it returns a result. However, when the `@Asynchronous` annotation is used, Fault Tolerance gets in between. When the method is called, Fault Tolerance first schedules the method to run later on a different thread, then it returns a `CompletionStage` or `Future`. This isn't the `CompletionStage` or `Future` that's returned from the method, because the method hasn't run yet, it's an object created entirely by Fault Tolerance. 
+
+image::/img/blog/FT-basic-asynchronous-execution.png[Fault Tolerance asynchronous execution flow]
+
+Sometime later, the method runs and returns a `CompletionStage` or `Future`. Now, Fault Tolerance will go and complete the `CompletionStage` or `Future` object which was returned to the caller earlier, with the value that was returned from the method. You might be wondering *_"What if the returned result is an exception? What can I do?"_*.  Read along.
```
Charlotte-Holt: >Now, Fault Tolerance will go and complete the `CompletionStage` or `Future` object which was returned to the caller earlier, with the value that was returned from the method.  

To

>Now, Fault Tolerance completes the `CompletionStage` or `Future` object that was returned to the earlier, with the value that was returned from the method.

Also, add a comma after "wondering", and there's an extra space before "Read along."

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+== `@Asynchronous` Use cases
```
Charlotte-Holt: Use doesn't need to be capitalized in this heading

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+What happens when the `serviceA()` method is called?
+
+Well, usually when a method is called it returns a result. However, when the `@Asynchronous` annotation is used, Fault Tolerance gets in between. When the method is called, Fault Tolerance first schedules the method to run later on a different thread, then it returns a `CompletionStage` or `Future`. This isn't the `CompletionStage` or `Future` that's returned from the method, because the method hasn't run yet, it's an object created entirely by Fault Tolerance. 
+
+image::/img/blog/FT-basic-asynchronous-execution.png[Fault Tolerance asynchronous execution flow]
+
+Sometime later, the method runs and returns a `CompletionStage` or `Future`. Now, Fault Tolerance will go and complete the `CompletionStage` or `Future` object which was returned to the caller earlier, with the value that was returned from the method. You might be wondering *_"What if the returned result is an exception? What can I do?"_*.  Read along.
+
+== `@Asynchronous` Use cases 
+So `@Asynchronous` sounds great, right? But how can you use it to make your asynchronous programs resilient? Let's go through 2 use cases:
```
Charlotte-Holt: I'd write out "two" in this sentence

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+1. <<Applying-to-asynch-api, Applying Fault Tolerance to an API call which is already asynchronous>>
```
Charlotte-Holt: >Applying Fault Tolerance to an API call which is already asynchronous

To

>Applying Fault Tolerance to an API call that is already asynchronous

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+2. <<Running-methods-in-parallel, Using the `@Asynchronous` annotation to do multiple things in parallel with resilience>>
+
+[#Applying-to-asynch-api]
+=== 1. Applying Fault Tolerance to an asynchronous API call
+One use of `@Asynchronous` is to apply Fault Tolerance to an asynchronous API call which returns a `CompletionStage`. Without `@Asynchronous`, you wouldn't be able to apply Fault Tolerance to the call. Let's see why.
```
Charlotte-Holt: Add a comma before "which returns a `CompletionStage`."

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+In this example, we'll use the JAX-RS client API for calling remote REST services, introduced in JAX-RS 2.1. We can build up a request to fetch a String from a given URL with a GET request and the return type is a `CompletionStage` of String:
```
Charlotte-Holt: >In this example, we'll use the JAX-RS client API for calling remote REST services, introduced in JAX-RS 2.1. We can build up a request to fetch a String from a given URL with a GET request and the return type is a `CompletionStage` of String:

To

>In this example, we'll use the JAX-RS client API, which was introduced in JAX-RS 2.1, for calling remote REST services. We can build up a request to fetch a String from a given URL with a GET request, and the return type is a `CompletionStage` of String:

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+If we call the `clientDemo()` method, without any annotations, it works as we expect. We call it, receive a `CompletionStage<String>` (named `response` in the example), and then we can add an action to take when the `CompletionStage` completes:
+
+[source,java]
+----
+response.thenAccept(System.out::println);
+--> responseText
+----
+
+However, what happens if we want to use the `@Retry` annotation to retry any failed requests with our method? The answer is that the method will not be retried.
```
Charlotte-Holt: I'd take out "The answer is that"

Charlotte-Holt (commented):
@Joseph-Cass Just a few suggestions from an initial look. Can you open a PR for the post to the `posts` dir on the `draft` branch?

Charlotte-Holt (approved):

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: I agree with the rewording, but personally prefer semi colons to dashes

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: I think the centring works well here, but have a look at the generated page now it's up. Happy to change it if you think it looks better left-aligned.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: This sentence is trying to say:

"Let's show you what using @Asynchronous does to your code when it's used and how it does it"

Though I realise the current wording is a little ambiguous. 

I think "can do" suggest something else, and doesn't reflect what follows.

drafts/Asynchronous-programming-made-resilient-with-MPFT.adoc
Joseph-Cass: I agree with the first sentence. 

The second sentence should be: "We can build up a request to fetch a String from a given URL with a GET request, where the return type is a `CompletionStage` of String

PR #363, opened 2020-05-12
Delete 2019-11-08-Diagnose-slow-requests-more-easily-with-request-log…

steven1046:
…ging.adoc

https://github.com/OpenLiberty/blogs/blob/master/drafts/2019-11-08-Diagnose-slow-requests-more-easily-with-request-logging.adoc is a duplicate of https://github.com/OpenLiberty/blogs/blob/master/publish/2019-11-08-diagnose-slow-requests-easily-190011.adoc which is already published so the draft one could probably be deleted.


lauracowen (approved):
thx

PR #424, opened 2020-05-22
GraphQL - minor edits for style and typos

lauracowen:
I've also switched round some of the sentences in the opening paragraphs to make it clear from the start what the benefit of MP GQL is. I've relegated info about the spec and feature name to the end of the intro as that's not so interesting, though useful to some.

Otherwise, I think I've only made minor typo/punctuation/grammar fixes.


andymc12 (approved):
All of the edits look good.  Thanks Laura!

PR #439, opened 2020-06-01
Update the description of what the Asynchronous annotation does

Azquelt:
The second point was describing something you could usefully do with
`@Asynchronous`, and it wasn't referenced at all in the following text.

Update the second point so that it describes what FT actually does when
`@Asynchronous` is present on a method and add detail later in the section
explaining this in more detail.

The following section entirely covers the use case which was described
by the old text for point 2.

### DIFF
--- a/posts/2020-06-03-asynchronous-programming-made-resilient-with-mpft.adoc
+++ b/posts/2020-06-03-asynchronous-programming-made-resilient-with-mpft.adoc
@@ -30,7 +30,7 @@ But there's one, in particular, that's going to help you make your asynchronous 
 Adding the `@Asynchronous` annotation to a method does two things:
 
 1. Runs the method asynchronously
-2. Allows Fault Tolerance strategies to be applied to asynchronous methods
+2. If the method returns a `CompletionStage`, applies other Fault Tolerance strategies to the result of the `CompletionStage`, rather than to the result of the method
 
 Adding the `@Asynchronous` annotation to a method makes the method run asynchronously, meaning that rather than the annotated method running straight away on the main thread when it's called, it will run sometime later, usually on another thread. To use it, you need to add the `@Asynchronous` annotation to a method, and it must return either a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[`CompletionStage`] or a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html[`Future`]. You cannot apply Fault Tolerance to a returned `Future` (see why <<CS-preference, here>>), so a `CompletionStage` is preferred.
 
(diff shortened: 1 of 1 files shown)


posts/2020-06-03-asynchronous-programming-made-resilient-with-mpft.adoc
```diff
@@ -30,7 +30,7 @@ But there's one, in particular, that's going to help you make your asynchronous
 Adding the `@Asynchronous` annotation to a method does two things:
 
 1. Runs the method asynchronously
-2. Allows Fault Tolerance strategies to be applied to asynchronous methods
+2. If the method returns a `CompletionStage`, applies other fault tolerance strategies to the result of the `CompletionStage`, rather than the result of the method.
```
Charlotte-Holt: Suggestion

> If the method returns a `CompletionStage`, applies other Fault Tolerance strategies to the result of the `CompletionStage`, rather than to the result of the method

posts/2020-06-03-asynchronous-programming-made-resilient-with-mpft.adoc
```diff
@@ -53,7 +53,10 @@ Usually when a method is called, it returns a result. However, when the `@Asynch
 image::/img/blog/FT-basic-asynchronous-execution.png[Fault Tolerance asynchronous execution flow]
 
 Sometime later, the method runs and returns a `CompletionStage` or `Future`.
-Now, Fault Tolerance completes the `CompletionStage` or `Future` that was returned earlier, with the value that was returned from the method. You might be wondering, *_"What if the returned result is an exception? What can I do?"_*. Read along.
+
+If the method returned a `Future`, then the last step is for Fault Tolerance to update the `Future` it returned to the user earlier so that it delegates to the `Future` returned from the method, allowing the user to see the result from the method.
```
Charlotte-Holt: Suggestion

>If the method returns a `Future`, then the last step is for Fault Tolerance to update the `Future`  it previously returned to the user so that it delegates to the `Future` returned from the method, allowing the user to see the result from the method.

posts/2020-06-03-asynchronous-programming-made-resilient-with-mpft.adoc
```diff
@@ -53,7 +53,10 @@ Usually when a method is called, it returns a result. However, when the `@Asynch
...
+
+However, if the method returned a `CompletionStage`, Fault Tolerance will wait for the `CompletionStage` to complete, before completing the `CompletionStage` it returned to the user earlier. Additionally, if the `CompletionStage` from the method completes with an _exception_, then other Fault Tolerance policies will be applied, just as if that exception was thrown from the method body itself. This last step is cruicial for allowing Fault Tolerance to work with other APIs which return a `CompletionStage`, as we'll see in the next section.
```
Charlotte-Holt: Suggestion

>However, if the method returns a `CompletionStage`, Fault Tolerance waits for the `CompletionStage` to complete, before completing the `CompletionStage` it previously returned to the user. Additionally, if the `CompletionStage` from the method completes with an _exception_, then other Fault Tolerance policies are applied, just as if that exception was thrown from the method body itself. This last step is crucial for allowing Fault Tolerance to work with other APIs that return a `CompletionStage`, as we'll see in the next section.

Charlotte-Holt (commented):
@Azquelt - Just a few small suggestions

Joseph-Cass:
I agree that the second point of "What does Asynchronous do?" was not properly addressed in that section, so I think the additional paragraph works well

Joseph-Cass (dismissed):
I agree that the second point of "What does Asynchronous do?" was not properly addressed in that section, so I think the additional paragraph works well.

I also agree with all of @Charlotte-Holt's suggestions.

Azquelt:
I've applied @Charlotte-Holt's suggestions

posts/2020-06-03-asynchronous-programming-made-resilient-with-mpft.adoc
```diff
@@ -53,7 +53,10 @@ Usually when a method is called, it returns a result. However, when the `@Asynch
 image::/img/blog/FT-basic-asynchronous-execution.png[Fault Tolerance asynchronous execution flow]
 
 Sometime later, the method runs and returns a `CompletionStage` or `Future`.
-Now, Fault Tolerance completes the `CompletionStage` or `Future` that was returned earlier, with the value that was returned from the method. You might be wondering, *_"What if the returned result is an exception? What can I do?"_*. Read along.
+
+If the method returns a `Future`, then the last step is for Fault Tolerance to update the `Future`  it previously returned to the user so that it delegates to the `Future` returned from the method, allowing the user to see the result from the method.
```
Joseph-Cass: Think there's a double space between "the `Future`" and "it previously"

Joseph-Cass:
```suggestion
If the method returns a `Future`, then the last step is for Fault Tolerance to update the `Future` it previously returned to the user so that it delegates to the `Future` returned from the method, allowing the user to see the result from the method.
```

Azquelt: fixed

Joseph-Cass (approved):
LGTM :)

PR #583, opened 2020-07-15
Merge final blog to draft

Charlotte-Holt:


lauracowen (approved):
ok, almost all the deletions are fine (clean-up we should've done ages ago). The couple we need, I've taken copies of or we have copies in other branches so I'll PR it back when we've merged this.

PR #594, opened 2020-07-21
Updated to remove support option that no longer exists

NottyCode:
Update the november post on support options to remove the option that no longer exists.


lauracowen (changes requested):
Can you change "you can either run" to "you can run either".

lauracowen (dismissed):
LGTM

lauracowen (approved):

PR #648, opened 2020-08-14
Third party microsoft ol

jakub-pomykala:


lauracowen (approved):
tags file is fine - draft needs updating with it from time-to-time
Not sure why my beta blog posts are showing up here becuase they should already be in `draft` but it's fine to merge them anyway.

PR #666, opened 2020-08-20
Process instructions updates suggested by EricL

lauracowen:


README.md
```diff
@@ -53,6 +53,9 @@ These steps are to be completed by the author of the blog post.
 5. Push the file to GitHub, then create a pull request (PR) into the `draft` branch.
 
    Link the PR to the issue you created in Step 1.
+   Anyone can review the PR but @lauracowen (or other admin) need to merge it.
```
ericglau, line 56: Is this statement true?  So far, I've seen that anyone can review the PR and then the originator can merge it to `draft`, which seems fine.  It seems unnecessary to have an admin merge it to `draft`.

lauracowen: oh if that's the case (I can't tell), then that's perfect (and makes more sense). I'll just modify that text. Thanks.

PR #716, opened 2020-09-02
Publish Jaeger blog post - Wed 16th Sept

lauracowen:
Publishing date 16th Sept


lauracowen:
@fwji 
I've edited it to be shorter and more focused. Part of what I've taken out is where you've given the reader options - there's not space in a blog post to be full documentation (and it's not our product to document anyway). Can you take a look and see what you think, please? If you'd like to add in more links to Jaeger etc's documentation, feel free. I've made some edits for style so if you do any more changes, make sure you git pull `jaeger-branch` before you do.

See https://draft-openlibertyio.mybluemix.net/blog/2020/09/03/opentracing-jaeger-openshift.html

- [ ] "The best way to deploy Jaeger on OpenShift is using the Jaeger Operator." - according to Red Hat, it's not: https://docs.openshift.com/container-platform/4.5/jaeger/jaeger_install/rhbjaeger-installation.html - they imply the best way is to just install Service Mesh which includes Jaeger. I think we need to be clearer here. I know this post isn't about doing the all-in observability but I don't think saying "the best way" without qualification works.
- [ ] The post provides the command needed to create the `jaegertracing` project but then tells them to just create the instance using the yaml - this is inconsistent. Either we should provide a link to where they can find out how to do that (https://docs.openshift.com/container-platform/4.5/jaeger/jaeger_install/rhbjaeger-deploying.html) or we just provide the commands inline. As this post has been set up as providing a tutorial, I've added in what I think are the two commands needed. Please check.
- [ ] Can you get someone to test this post, please? And then approve it in this PR. It doesn't need to contain every detailed step (it's reasonable to expect the reader to, say, know how to use OperatorHub in this case) but it needs to be reasonable expectations (eg providing the command for creating the project but then not providing the commands to create the Jaeger instance doesn't seem very consistent - should they know this or not?).

fwji (dismissed):
LGTM!

I have made minor changes based on your comments:

1. Instead of saying the best way, I changed to the foolproof way to install Jaeger on OpenShift
2. You added command looks good, however the second command you added is exactly what I did in my next command block, so I removed it for redudancy reason.

tonyreigns (dismissed):
LGTM

fwji (approved):

PR #719, opened 2020-09-03
Microsoft azure blog

jakub-pomykala:


lauracowen (changes requested):
Thanks, some minor style things - sorry I didn't do them inline (hope you can work out where they apply):
- "Instead, contemporary applications leverage these functions (Identity and Access Management or IAM for short) from an external provider." to "Instead, contemporary applications make use of Identity and Access Management (IAM) services from external providers."
- "Liberty social login feature" to "Liberty Social Media Login feature"
- "example on" to "example of" (couple of instances)
- (heading) "Configure social login as OpenID Connect client" to "Configure social media login as OpenID Connect client"
- put "socialLogin-1.0" in backticks (monospace)
- "Above code sample in GitHub repository" to "(Find this code sample in GitHub.)" (make "code sample" into the link and the rest normal text; make sure it ends with a full-stop).
- put "oidcLogin" in backticks (it's a config XML element)
- "userNameAttribute" in backticks
- "There are a number of Azure AD token values you can use that are listed here." put the link on "a number of Azure AD token values" then remove "that are listed here"
- "preferred_username", "oid" in backticks.
- "defaultSSLConfig" and "trustDefaultCerts" and "true" in backticks
- "that only users with the role users can access." put the second "users" (but not the first) in backticks)
- Change the link text below the config example in the same way as for the first.
- Remove "Picture 1:" from the picture caption.
- For each example code, can you re-word the sentence beneath as for the first one?
- "The CafeResource implements" to "The CafeResource service implements"
- "groups claim" or "groups claims" - put "groups" in backticks (a few instances)
- "you will need to " to "you need to "
- "type as ‘Security’" - remove the quotation marks and put "Security" in backticks
- "you will also need to: find" to "you also need to find"
- "‘Token configuration’ > select ‘Add groups claim’ > select ‘Security groups’ " - "Token configuration > Add groups claim  > Security groups" then just put the whole thing in bold (ie ** at start and end of the whole lot).
- "ID token > expand ‘ID’ and select ‘Group ID’ in ‘Customize token properties by type’ " - similarly, remove quotation marks, replace "and select" with " > ", remove "in" then put the whole lot in bold.

lauracowen (approved):
thank you

PR #763, opened 2020-09-22
Fix 20.0.0.9 bugs link

mbroz2:
Current link is pointing to 20.0.0.7 instead of 20.0.0.9


lauracowen (approved):
Thanks. I've also fixed the "JJAX-RS" typo Jared raised.

PR #785, opened 2020-09-24
Publish 50 guides post and tomorrow's 20.0.0.10 release blog post

lauracowen:


lauracowen:
Also connects to #785

PR #800, opened 2020-10-08
Publish blog post for MP language support 

Charlotte-Holt:


lauracowen (approved):
Forgot these changes were waiting to be published. It's a mass update of all the old blog posts so that they all use attributes consistently. They look fine to me.

PR #818, opened 2020-10-20
Try it now section for All-beta

jakub-pomykala:
Seems like I have copied the description from the try it now section for Jakarta EE 9 over to the All Beta section instead of adding the right description


lauracowen (approved):
LGTM

PR #824, opened 2020-10-20
OCP Log Forwarding updates

Channyboy:
Update author information, update file name with newer date so it doesn't appear as an _old_ post.


fmhwong (approved):
LGTM

Issue #868, opened 2020-11-13
Blog post for getting started with java-openliberty odo/devfile stack


scottkurz:
1. Take simple intro app, clone, create java-openliberty component, push
 * make iterative change, 

2. Take existing guide app, create java-openliberty component, push
* ?maybe make change

Question:
 * deploy just to OCP or minikube/generic k8s?
---
References
* odo.dev, 
* application-stack/wiki

* CP4A 4.3 Kctr????  (Or keep this scoped to odo)

yeekangc:
Perhaps we can consider a video (as a demo) along with this too (that can be linked to in the blog post).

scottkurz:
First pass at:  https://github.com/OpenLiberty/blogs/blob/ol-stack/posts/2020-11-30-open-liberty-devfile-stack.adoc

awisniew90:
@GraceJansen  - draft PR has been merged with a technical review.

gljrobinson:
@Charlotte-Holt and @dmuelle ready for editing whenever you get the chance.

Charlotte-Holt:
ID review updates incorporated at https://blogs-draft-openlibertyio.mybluemix.net/blog/2021/01/20/open-liberty-devfile-stack.html.

awisniew90:
@Charlotte-Holt - LGTM!

Charlotte-Holt:
This published at https://openliberty.io/blog/2021/01/20/open-liberty-devfile-stack.html

PR #949, opened 2020-12-14
Update OL post in draft with screen shots

awisniew90:


awisniew90:
Fixes #868

awisniew90:
@GraceJansen

PR #963, opened 2020-12-30
Small update to OCP Log Fowarding blog

Channyboy:
Fix value that is to be expected in the Fluentd config map.
And a note indicating what the value would be instead in OCP 4.3 to 4.5


posts/2021-01-06-ocp-log-forwarding.adoc
```diff
@@ -0,0 +1,425 @@
...
+. To check whether the logs are forwarded to the specified outputs, run the following command:
++
+[source]
+----
+[root@ocp ~]# oc -n openshift-logging get cm fluentd -o json | jq -r '.data."fluent.conf"' > fluentd-with-logfowarding.conf
+----
++
+This command displays the ConfigMap configuration for Fluentd inside OpenShift. Check whether the outputs are defined inside the configuration file.
+
+* In the following configuration file example, you can see the `elasticsearch` and `fluent-forward` outputs that are referenced by `@label`:
```
dmuelle: The intro sentence here still makes a reference to the `elasticsearch` output, which is only valid for 4.3-.46, and has been removed from the examples. Suggest the following edit:

In the following ConfigMap file example, the `FLUENTD_FORWARD`  output type is specified for the `CONTAINER_LOGS` pipeline:

I don't think we need to mention DEFAULT here, as it's not the focus. However, want to check-  In our yaml example, it says 

`disableDefaultForwarding: true` 

is the default output still available in that case? Probably ok but just want to make sure the examples mixing the default and custom outputs are viable.

posts/2021-01-06-ocp-log-forwarding.adoc
```diff
@@ -0,0 +1,425 @@
...
+    <store>
+      @type relabel
+      @label @FLUENTD_FORWARD
+    </store>
+  </match>
+</label>
+...
+```
++
+**OpenShift 4.3 to 4.5 only:**  The value `@DEFAULT` would be `@ELASTICSEARCH`
```
dmuelle: suggested edit:

**OpenShift 4.3 to 4.5 only:**  The `@label @DEFAULT` value is replaced by `@label @ELASTICSEARCH`.

dmuelle (changes requested):
couple minor changes and a quick question

posts/2021-01-06-ocp-log-forwarding.adoc
Channyboy: A good catch on the `disableDefaultForwarding: true` value.
This is a relic from the 4.3-4.5 configurations. It would've disabled the _default_ behavior and that's why I assume the need to define Elasticsearch output in those versions. However it would seem that  attribute is not recognized anymore in 4.6 and nothing _happened_ when I was running through.

posts/2021-01-06-ocp-log-forwarding.adoc
```diff
@@ -0,0 +1,424 @@
...
+
+
+. Create the `log-forward-instance.yaml` file. Configure the outputs and pipelines.
++
+* The `outputs` section of the file specifies where to send the logs. To establish a secure connection to an output, you must specify a secret, such as the one that you created in <<keycert-secret,step 2b of the *Before you begin* section>>. If you want to use an insecure connection for test purposes, you can omit the secret from your output configuration.
+* The `pipelines` section of the file specifies which log types are sent to the configured outputs.
++
+The following sample `log-forward-instance.yaml` file defines two `outputs`: a secure connection to a Fluentd server named `fluentd-server-secure` and an insecure connection to another Fluentd server named `fluentd-server-insecure`. Note the different protocols that are defined for the _url_ in each output (tls vs tcp). Remove the output type that you do not want to use in your file. Each log type and its corresponding output reference is defined in the `pipelines` section. The `default` output reference points to the internal OpenShift Container Platform Elasticsearch instance:
++
+```
```
dmuelle, line 69: should one of the pipelines (`infra-logs` or `audit-logs`) reference the `fluentd-server-insecure` output? We've defined that output in the spec but not referenced it in an actual pipeline. The last two pipelines only reference the `default` output, which isn't really the focus here- should one of them also reference `fluentd-server-insecure` for demonstrative value?

dmuelle (dismissed):
changes look good- approved but see my comment on the example yaml file. Up to you as to whether it needs another edit

dmuelle (approved):
LGTM

PR #964, opened 2021-01-05
remove tags for ocp-log-forwarding

dmuelle:
remove tags since post is not ready to publish


Charlotte-Holt (approved):

mbroz2:
fyi, it's ok if tags go to staging/prod before a blog is published as it won't break/negatively impact anything.

PR #980, opened 2021-01-12
Remove redhat support staging

steven1046:


mbroz2 (changes requested):
Can you close this PR and instead do staging into prod?

PR #1006, opened 2021-01-19
Ga release blog issue edit

jakub-pomykala:
Removing Red Hat section from the GA release issue as we no longer require it.


gljrobinson (approved):
LGTM

PR #1034, opened 2021-01-26
Changing "messaging 3.0" to "Jakarta Messaging 3.0"

jakub-pomykala:
@dazavala has noticed that the heading is wrong. It is supposed to be "Jakarta Messaging 3.0"


dazavala (approved):
Thanks, @jakub-pomykala.  Approved.

mbroz2 (approved):

PR #1053, opened 2021-01-29
Add blog redirects file, moved from the ui repo

steven1046:
The build will now copy over and use this redirect file, so the blogs team can better control the old/new url mappings for blogs.

### DIFF
--- a/blog-redirects.properties
+++ b/blog-redirects.properties
@@ -1,0 +1,11 @@
+# Blogs redirects
+/news/=/blog/
+/news=/blog/
+/news/*=/blog/
+/blogs=/blog/
+/blogs/=/blog/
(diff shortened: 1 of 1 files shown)


gljrobinson:
@steven1046 is this issue complete?

steven1046:
Yes

gljrobinson:
Do we need this to be approved and merged by someone else?

steven1046:
Anyone can review it, the redirects still work in https://github.com/OpenLiberty/openliberty.io/blob/prod/src/main/webapp/WEB-INF/ui-redirects.properties and once this is merged in and makes it to prod we can remove those redirects from the UI repo.

dmuelle (approved):
LGTM

PR #1110, opened 2021-02-19
Adding lines and fixing sig bug fix section

jakub-pomykala:
Made the sig bug fix section inline, and also added lines in between sections (''')


austin0 (approved):
LGTM

PR #1130, opened 2021-02-23
Updating video embed code

andymc12:
Using different video embed code. Draft server showed "This content is blocked. Contact the site owner to fix the issue." instead of the embedded video.


jakub-pomykala (approved):
LGTM

PR #1162, opened 2021-03-01
Switch gradle group to `compileOnly`

andymc12:
Fixed minor bug discovered by @Joseph-Cass - the Gradle group should be `compileOnly` to avoid pulling the MicroProfile Rest Client APIs into the application when they are provided by the Liberty server.


Azquelt:
Exactly what the sample you had does will depend on what else is in your gradle file. With nothing else you'll get an error since there's no configuration defined for `mpRestClient`. If this came from a working sample, then I assume that elsewhere in the build.gradle file, it declared an `mpRestClient` configuration and used it for something.

I think `compileOnly` is what users should be using if they have a simple build.gradle and want to add a dependency on any of the microprofile APIs.

Azquelt (approved):

jakub-pomykala (approved):
LGTM!

PR #1245, opened 2021-03-23
21003 anchor links

dmuelle:


jakub-pomykala:
I have tested the `opentracing` anchor on the draft site and it works as expected.

jakub-pomykala (approved):
LGTM

Issue #1168, opened 2021-03-03
MP JWT 1.2 for 21.0.0.3


Emily-Jiang:
Blog post for MP JWT 1.2 @teddyjtorres please provide details

The MP JWT 1.2 blog post will outline the new MP Config properties that were added to the spec, support for JSON Web Encryption (JWE) tokens, and enhanced algorithm support. The blog will include usage examples for the new functionality as well.

MP 4.0 blog posts are expected to be done by **March 12**.

The Open Liberty release notes are presented in three places:
- [Open Liberty blog](https://openliberty.io/blog/) (GA and beta features; [example](https://openliberty.io/blog/2020/05/07/EJB-persistent-timers-20005.html))
- [Red Hat Runtimes release notes for Open LIberty](https://access.redhat.com/documentation/en-us/open_liberty/2020/) (GA features only; [example](https://access.redhat.com/documentation/en-us/open_liberty/2020/html/release_notes_for_open_liberty_20.0.0.5_on_red_hat_openshift_container_platform/features))
- [Red Hat Developer blog](https://developers.redhat.com/blog/) (GA features only; [example](https://developers.redhat.com/blog/2020/05/13/open-liberty-20-0-0-5-brings-updates-to-ejb-persistent-timers-coordination-and-failover-across-members/))

## 1. Open Liberty release blog post

The Open Liberty release blog post is written in asciidoc and pushed to the [OpenLiberty/blogs](https://github.com/openliberty/blogs) repo. The repo README contains [instructions on how to create, build, and review](https://github.com/OpenLiberty/blogs/blob/prod/README.md) the post, including a template file to help you.

Follow the instructions in the README and the template file to complete the following tasks:

- [x] Create a draft release blog post in asciidoc, build to [the draft site](https://draft-openlibertyio.mybluemix.net/blog/), and create a PR to the `staging` branch with a link to the post on [the draft site](https://draft-openlibertyio.mybluemix.net/blog/). Add all the reviewers (see below) to the PR so that they can add all their review feedback and approvals in the PR.
- [ ] Get the draft post reviewed by:
  - [x] All the people who contributed the content to the blog post.
  - [ ] At least one of the following people: @mbroz2
- [ ] Agree with @mbroz2 which feature will lead this blog post, then write a title, slug, summary first paragraph, and SEO front matter appropriately.
- [ ] Get the post approved by @mbroz2.
- [ ] On release day (usually a Friday) @mbroz2 will publish the post.

[...]

PR #1262, opened 2021-03-24
Add MP JWT 1.2 standalone blog to staging

ayoho:
For #1168

Draft site: https://draft-openlibertyio.mybluemix.net/blog/2021/03/26/MP-JWT-1.2.html

image
image
image
image
image
image
image
image
image


posts/2021-03-26-MP-JWT-1.2.adoc
```diff
@@ -0,0 +1,290 @@
...
+The `mp.jwt.verify.audiences` property allows you to create list of allowable audience values. At least one of these must be found in the `"aud"` claim of the JWT. Previously, this had to be configured in the server.xml file. Now you can configure the audiences in the MicroProfile Config property as follows:
+[source]
+----
+mp.jwt.verify.audiences=conferenceService, adminService
+----
+
+[#mp-jwt-decrypt-key-location]
+=== mp.jwt.decrypt.key.location
+
+To support JSON Web Encryption tokens, the `mp.jwt.decrypt.key.location` property allows you to specify the location of the Key Management Key. This is the private key that is used to decrypt the Content Encryption Key, which is then used to decrypt the JWE ciphertext. The private key must correspond to the public key that is used to encrypt the Content Encryption Key.
```
mbroz2, line 78: shouldn't "Key Management Key" be "Private Management Key"

mbroz2 (changes requested):

posts/2021-03-26-MP-JWT-1.2.adoc
teddyjtorres: "Key Management Key" is a technical term derived from the JWE specification and we want to be specific as to what Key, from the specification, this is referring to.

This is what the MP JWT 1.2 specification refers to in https://download.eclipse.org/microprofile/microprofile-jwt-auth-1.2/microprofile-jwt-auth-spec-1.2.html,

> "Key management key - public RSA key which is used to encrypt a content encryption key. mp.jwt.decrypt.key.location must point to a private RSA key matching this key."

ayoho: "Key management key" is the noun phrase we've used throughout our docs. The [JWE spec](https://tools.ietf.org/html/rfc7516#section-2) defines the phrase "key management mode" as the "method of determining the Content Encryption Key value to use". Building off of that, we decided the key used to wrap the Content Encryption Key would then be the key management key. Our `<jwtBuilder>` and `<mpJwt>` elements have corresponding `keyManagementKeyAlias` attributes.

mbroz2: ah, thanks for the explanation.  Key Management Key sounded odd to me, like a mistake, and when I searched for jwt.decrypt.key.location, I saw that smallrye seemed to refer to it as the "Private Decryption Key" and in an [MP issue](https://github.com/eclipse/microprofile-jwt-auth/issues/193) it was also referenced to as Private Decryption Key.

mbroz2 (approved):

PR #1315, opened 2021-04-08
Admincenter blog 

aknguyen7:
https://draft-openlibertyio.mybluemix.net/blog/2021/04/15/admin-center-21004.html

image

### DIFF
--- a/posts/2021-04-15-admin-center-21004.adoc
+++ b/posts/2021-04-15-admin-center-21004.adoc
@@ -54,80 +54,37 @@ server create myServer
 [start=3]
 . Configure the `server.xml` file to enable Admin Center and secure login.
 +
-[loweralpha, start=a]
-.. Enable the Admin Center feature.
-+
-[source, xml]
+To use Admin Center, you must enable the `adminCenter-1.0` feature and https://www.openliberty.io/docs/latest/reference/feature/appSecurity-3.0.html#_configure_rest_api_access_roles[configure access roles] for at least one user to be in either the `reader-role` or `administrator-role` security role. You can use the following example `server.xml` file to try out Admin Center. This example enables the `adminCenter-1.0` feature and configures a https://www.openliberty.io/docs/latest/user-registries-application-security.html#_basic_user_registries_for_application_development[basic user registry] that defines users in the `reader-role` and `administrator-role` security roles:
+
+[source,xml]
 ----
-  <featureManager>
-    <feature>adminCenter-1.0</feature>
-  </featureManager>
+<server description="Admin Center example server">
+   <!-- Enable features -->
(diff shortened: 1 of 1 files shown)


dmuelle:
Hi @aknguyen7,  post looks good!  I worked through it and tried out the admin center feature. For the most part I think the info is straightforward and easy to follow. I've finished my initial round of ID edits and rebased the PR to draft so we can deliver them there first. Once we have the ID review sorted, we can change the filename to reflect the publish date (either Wed of Fri, depending on approvals) and open a PR to staging.

When you have a chance, can you look over the updated PR and let me know if the changes look ok? Mostly just minor edits per ID guidelines for style, clarity and concision. Just had an couple questions:

- Are Java Batch and OpenID Connect the only features with dedicated tools in Admin Center?

- From a content design perspective, I feel like there's too much setup. It's helpful if someone is starting from 0 with Liberty, but for users with even a little experience, it makes the task seem more complicated than it actually is. I found myself skipping through the entire setup section and just copying the example server.xml- especially since the sample has inline comments.

I think we could reduce the set up section to a single step, possibly two, that just says, here's what you need in your server.xml to run admin center, and here's an example you can use to try it out. I'm happy to take a stab at revising that section but just wanted to check with you before I made any structural changes. Let me know what you think. Thanks!

aknguyen7:
@dmuelle 
1. The Java Batch and OpenID Connect tools are the features that can be enable in Admin Center Open Liberty (by default the Admin Center) include the Explore, Server Config tools. 
2. I'm thinking we can mention that if user would like to skip the details setup they can jump to the session to just copy the entire server.xml?

mbroz2:
Let me clarify regarding the Admin Center tools in Open Liberty:
"Explore" & "Server Config" are provided by the Admin Center feature.  In other words, they're available as soon as you enable Admin Center.
"Java Batch" is an Admin Center tool that will appear if the user also enables the batchManagement-1.0 feature.
The 3 OIDC tools are actually standalone UIs, which can be accessed directly without the Admin Center feature, as long as the openidConnectServer-1.0 feature is enabled.  However, if both features are enabled, they can also be accessed via Admin Center (and if recall correctly, is also the only way to set a11y rules)

I had the same thought regarding the setup process; it makes it feel daunting as written because it goes into a bunch of details.  That's why for the release blog, I instead just included something much shorter and the smallest server.xml necessary to run Admin Center.  I think we should provide/lead with a similar option in the admin center blog.  Perhaps we can move the server.xml snippet to the top of the setup and say something like "To get started right away, just copy the the configuration and start your server" and either then provide all the extra info regarding details for the setup, move them to the bottom of the blog, or trim them down.

dmuelle (approved):

PR #1504, opened 2021-06-16
Finding fixpacks new

dmuelle:


barbj:
Hi @dmuelle 

bullet under method1:

_This label indicates the first fix pack that addresses the issue. However, if the release bug label was added to an issue before its first fix pack was delivered, the release:yynnn label might not appear on the issue._

Should be changed to

This label indicates the first fix pack that addresses the issue. However, if the release bug label was added to an issue **_after_** its first fix pack was delivered, the release:yynnn label might not appear on the issue.


<hr/>

bullet under method2:

to
_If the GitHub issue has the release bug label, you might find the issue in the Fix list for IBM WebSphere Application Server Liberty - Continuous Delivery, which lists the Git issues that are addressed in each Liberty fix pack._

please add:

_The issue might not appear in the fix list if the release bug label was added to an issue after its first fix pack was delivered._

<hr/>

bullet under method3:

_If the issue does not have a release:yynnn **or release bug label**, you can use this method._

If you've gotten to method 3, and have already attempted methods 1 and 2 w/o success, I intend for you to try this method.

If the issue does not have release:yynnnn, AND the release bug label is on the issue, BUT it is not in the fix list, It is possible that someone added release bug after the fix list was created. 

Please either just remove release bug, or change it to _it is not in the fix list_.  If it doesn't have release bug, it's not going to be in the fix list, so you can probably leave it off.  

You can go around manually adding labels to issues whenever you want.  The data on the fixpack data on the PRs however is reliable.  That is one of the reasons why I go through the process of having you verify the data in in the release label against the PR in method 1.

<hr/>

When I wrote the document, the methods were all titled associated with the issues, which is what the document is about "Finding the Liberty fixpack from a GitHub issue":

- If the issue has the release bug label...
- Finding the fix pack for a 'release bug' issue
- Finding the fix pack for a non-'release bug' issue

Now we have this:

- Find the fix pack information in the release:yynnn issue label.
- Finding the fix pack information in the Liberty fix list.
- Finding the fix pack information in the gm-yynnn release tag

Those aren't all relating to the issue.  They're also abstract, requiring that you know sometimes multiple pieces of information in order to know to pick that item.

My intention was:

I'm in the issue my issue has release bug, what do I do?: **Finding the fix pack for a 'release bug' issue**

It doesn't have release bug, what do I do?: **Finding the fix pack for a non-'release bug' issue**

Ya, I walk you through it at the top, but at its core, its release bug and not release bug.

I thought that we got the write blog articles how people talk and naturally find information, not how you write/read technical documentation.  Am I wrong?

PR #1510, opened 2021-06-17
Finding edits

dmuelle:


barbj:
@dmuelle 
I got this one wrong in the original text, can you fix it?  I fixed it when editing your rewrite, but now that you've restored some stuff, my mistake was also restored!

**Under method2, change:**
If the release bug label was added to an issue `before` its first fix pack was delivered, the release:yynnn label might not appear on the issue.

-to-

If the release bug label was added to an issue `after` its first fix pack was delivered, the release:yynnn label might not appear on the issue.

Sorry about this...

dmuelle:
@barbj NP- fixed!

barbj (approved):

Issue #1781, opened 2021-10-04
Blog post for Philly JUG presentation


ajm01:
blog post to highlight a recent discussion I held with the Philadelphia JUG

dmuelle:
published at https://www.openliberty.io/blog/2021/10/26/openliberty-jug-presentation.html

PR #1842, opened 2021-10-25
minor edits

dmuelle:
#1781


dmuelle:
Hi @ajm01 - THanks for making the edits from the review- I noticed a few other very minor things after I merged your PR so I made these edits - if everything looks god to you we can merge this and then send it to staging. We'll need to change the filename to reflect the new publish date.

@GraceJansen when would you like to target this to publish?

ajm01 (approved):

PR #1866, opened 2021-11-01
update default value in beta blog

dmuelle:
it might be confusing to say the default value for `mp.openapi.extensions.liberty.merged.info` is "none", because the actual default value for `mp.openapi.extensions.liberty.merged.exclude` is `none`. So, in the table "none" is being used in two different ways, the first time as an actual value, then again to say that there is no default value, which is potentially confusing. This PR updates the table to use "N/A" instead of "none" for `mp.openapi.extensions.liberty.merged.info`


priyyanshi (approved):

priyyanshi:
LGTM.

PR #2002, opened 2022-01-20
STAGING - 22.0.0.2-beta blog

ryan-storey:


tevans78 (dismissed):

posts/2022-01-20-ssl-certificates.adoc
```diff
@@ -0,0 +1,100 @@
...
+* <<ssl, Support new change in the SSL certificate with urlencoded string>>
+
+[#ssl]
+=== Support new change in the SSL certificate with urlencoded string
+
+An X.509 certificate is a digital certificate that is used to manage identity and security in internet communications and computer networking.
+
+Intermediate servers, such as a web server, proxy server, or load balancer can send X.509 certificates to Open Liberty servers. Due to the deprecation of long lines folding in the `HTTP` header continuation, some open source intermediate servers might send a client certificate in the Privacy-Enhanced Mail (`PEM`) url-encoded format.
+
+Prior to Open Liberty 22.0.0.2-beta, this new url-encoded certificate was not supported. Any request with this new format was rejected and the request terminated.
```
dmuelle: sorry- should've caught this before- I think line 40 and line 60 should swap places. The important info that we want to highlight is that the beta adds this capability. As is, that's not explained until the end of the section. Better to highlight it here and then mention previous behavior at the end.

tevans78 (approved):

Issue #2065, opened 2022-03-03
Working with dev mode in Eclipse IDE


kathrynkodama:

yeekangc:
This has been published.  Closing.

PR #2067, opened 2022-03-09
Working with Liberty dev mode in the Eclipse IDE post

kathrynkodama:
#2065


posts/2022-03-11-dev-mode-with-eclipse.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+image::/img/blog/dev-mode-eclipse-ide-terminal.png[Starting Liberty dev mode through the Eclipse IDE terminal,width=70%,align="center"]
+
+== Running tests
+
+You can run tests on demand by pressing Enter in the terminal running dev mode.
+
+== Attaching a debugger
+
+You can attach a debugger to the running server to step through your code at any time. In Eclipse you can create a debug configuration by "Run" -> "Debug Configurations..." -> "Remove Java application". The default port for debugging is `7777`.
```
scottkurz: Remove -> Remote

posts/2022-03-11-dev-mode-with-eclipse.adoc
```diff
@@ -0,0 +1,95 @@
...
+    }
+    dependencies {
+        classpath 'io.openliberty.tools:liberty-gradle-plugin:3.3'
+    }
+}
+----
+
+== Starting dev mode
+
+Start dev mode through the Eclipse IDE terminal ("Window" -> "Show View" -> "Terminal") or through your operating system's terminal with the following commands:
```
scottkurz: When I just do this "Show view" way to access the terminal I don't get a command prompt.    OTOH, if I go to Package Explorer, right-click, "Show in Local Terminal" and then make a selection like "Terminal", I get a prompt right at the corresponding directory...which is more useful.

posts/2022-03-11-dev-mode-with-eclipse.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+You can quit dev mode at any time by pressing CTRL+C in the terminal, or type 'q' and press Enter.
+
+== Working with dev mode in the Eclipse IDE with Liberty Developer Tools installed
+
+If you have https://marketplace.eclipse.org/content/ibm-liberty-developer-tools[Liberty Developer Tools] installed in your Eclipse workspace, the server configuration editor views can be helpful as you modify your `server.xml` and other configuration files.
+
+image::/img/blog/liberty-dev-tools-server-config-view.png[Liberty Developer Tools server configuration view for the Eclipse IDE,width=70%,align="center"]
+
+Dev mode handles a lot of the heavy lifting for developers when using the Liberty runtime. With dev mode, you no longer need to manage your Liberty instances through the "Servers view" of Liberty Developer Tools.
```
scottkurz: "Servers view" => "Servers" view

posts/2022-03-11-dev-mode-with-eclipse.adoc
```diff
@@ -0,0 +1,95 @@
...
+Disable Liberty Developer Tools' integration for Maven and Gradle projects:
+
+* For Maven projects: Window -> Preferences -> Maven -> Liberty Integration to "No".
```
scottkurz: I think this is OK.   Another idea I've seen done would be to format with say bold all the things that are the menu, etc. selection text, e.g. more like **Window -> Preferences -> Maven -> Liberty Integration** drop-down to "No".   If you think it reads better you could change, not a big deal.   I know we already have several screenshots ...we could add one more here perhaps. Or not.

scottkurz (commented):
Hi Kathryn, I attached some minor comments here including a typo..

Besides that though, I think maybe this could use a bit more introduction.

The key insight here is to leverage the Liberty Tools server config editor while disabling the Maven integration.    

There are also a couple smaller details and examples on running dev mode. 

But dev mode is pretty easy to work with .. those details aren't maybe too critical since a good number of Eclipse/Maven users could probably figure that out.   But the "key insight" is maybe a bit buried in the mechanics and could surface to more of an intro level perhaps?

So the intro would be more like:  "use dev mode and still get the benefits of Liberty server config editor by disabling Liberty maven integration".  

If that review helps spur ideas great..if not let me know and I could take a stab at writing something.

posts/2022-03-11-dev-mode-with-eclipse.adoc
kathrynkodama: Thanks Scott, this is a good suggestion. I've made this update and added a new screenshot.

kathrynkodama:
> So the intro would be more like: "use dev mode and still get the benefits of Liberty server config editor by disabling Liberty maven integration".

Thank you @scottkurz! I see what you mean about refocusing the intro. I want to keep the Liberty Developer Tools plugin optional for users and emphasize that dev mode is the path forward. I've rephrased the intro to include mention of the Liberty Developer Tools plugin, and I've moved the "Working with dev mode in the Eclipse IDE with Liberty Developer Tools installed" section to the top of the post. Let me know what you think.

kathrynkodama:
cc: @yeekangc

scottkurz (approved):
I like the re-org.  Thanks for considering my comments.

gljrobinson (approved):
LGTM

PR #2069, opened 2022-03-09
Release template update and tags

mbroz2:
Include guides in release template

### DIFF
--- a/templates/ga-release-post.adoc
+++ b/templates/ga-release-post.adoc
@@ -69,6 +69,12 @@ In link:{url-about}[Open Liberty] RELEASE_VERSION:
 * <<TAG_3, FEATURE_3_HEADING>>
 * <<bugs, Notable bug fixes>>
 
+
+// // // // // // // //
+// If there were updates to guides since last release, keep the below, otherwise remove section.
+// // // // // // // //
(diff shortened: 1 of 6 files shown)


templates/ga-release-post.adoc
```diff
@@ -69,6 +69,12 @@ In link:{url-about}[Open Liberty] RELEASE_VERSION:
 * <<TAG_3, FEATURE_3_HEADING>>
 * <<bugs, Notable bug fixes>>
 
+
+// // // // // // // //
+// If there were updates to guides since last release, keep the below, otherwise remove section.
```
dmuelle: I know this is internal so maybe not a concern but should use "following" instead of "below" here and at 229 for accessibility reasons. Otherwise LGTM

dmuelle (dismissed):

templates/ga-release-post.adoc
mbroz2: No, you're right; I always make sure not to allow for any directional words in the actual blog posts, but here I just thoughtlessly followed the precedents in the template.  I'll update the PR to remove them from template(s)

mbroz2: Updated all templates to remove directional words.

ryan-storey:
LGTM

ryan-storey (approved):

PR #2095, opened 2022-03-16
Fix README Links to blogs-draft site

kinueng:


kinueng:
I have been noticing lately that GitHub UI is unable to automatically determine that the changes to `blog_tags.json` and `templates/beta-release-post.adoc` are already in `draft` even if the commit hash is different.  Maybe I am missing something but I tested doing this PR merge on my local machine and only saw a diff to the README.md.

dmuelle:
I see this issue and similar often in the docs as well. I don 't understand why it happens when it does but for me it has not been confined to changes made in the UI. My best guess is it has something to do with the fact that draft and prod are never really in sync, because we always have content in draft that isn't ready to go to prod. But after changes go from a working branch first to draft, and then to staging, we merge staging into prod, which might create some kind disconnect with draft, regardless of the fact that the content is the same in the affected topics. 

I really have no idea and just work around it though. From what I have seen, a few extra files being pulled is only a problem when the PR shows actual branch conflicts that have to be resolved. If it's just a couple files and no conflicts, I've never had a problem after merging it in. But if there are conflicts, or if a large amount of unrelated files are being pulled in, I have to debug and figure out how to get around it.

dmuelle (approved):
This is weird though because In thought these exact changes had already been "re-merged" to draft. But I dont think there's anything here that will cause a problem

PR #2192, opened 2022-04-11
Libertydevops 1 gitjenkinshelmiks

shamjithantholi:
please merge


yeekangc (commented):
Let us discuss further and please consider what is suggested in https://github.com/OpenLiberty/blogs/issues/2217.

yeekangc (changes requested):
Need to address what is noted in #2217 first.

lauracowen (approved):
I'm going to approve this into draft so that we can see it built. I'll then do an editorial review, which might still involve changes, but it'll be easier to follow than keep doing it by issue.

lauracowen:
I'm merging as most of the changes have been made and it'll be easier to review and edit if we get it built to draft.

PR #2275, opened 2022-05-25
add jenkins cic-cd post tags to staging

dmuelle:


kinueng (dismissed):
LGTM

mbroz2 (commented):
I think you got the key ones, but I would suggest also adding `test` tag (could also add `security` though that part is pretty sparse.....)

dmuelle:
updated to add security tag

gljrobinson (approved):
LGTM

PR #2295, opened 2022-06-01
Remove DZone and Travis mentions

lauracowen:
Travis mentions don't make sense if you're coming to it new. DZone step no longer needed.


lauracowen:
Sorry Kin, another one after finding out new info.

kinueng (approved):
LGTM

PR #2344, opened 2022-06-16
Lauracowen readme issue link update

lauracowen:
Just a minor change to direct reader to the right issue type for their context.


lauracowen:
pulled in the world when I didn't want it

PR #2354, opened 2022-06-20
Publish recent edits to README and Database TLS and buildpacks posts

dmuelle:
- https://blogs-staging-openlibertyio.mybluemix.net/blog/2021/06/04/database-ssl-primer.html
- https://blogs-staging-openlibertyio.mybluemix.net/blog/2022/04/01/cloud-native-liberty-buildpack.html
- https://github.com/OpenLiberty/blogs/tree/staging#writing-and-publishing-blog-posts-on-the-openlibertyio-blog

PR #2355, opened 2022-06-20
Update 2022-04-01-cloud-native-liberty-buildpack.adoc

dmuelle:
FIx per @mbroz2 suggestion in #2354


kevin-ortega (approved):
Agree with Michal's suggestions.

PR #2389, opened 2022-07-14
22008beta update

mbroz2:
Looks good on draft: https://draft-blogs-openlibertyio.mybluemix.net/blog/2022/07/07/separate-stack-trace-22008-beta.html
image


blog-redirects.properties
```diff
@@ -27,3 +27,4 @@
 /blog,/microprofile/2021/03/31/microprofile-config-2.0.html=/blog/2021/03/31/microprofile-config-2.0.html
```
kinueng, line 27:
```suggestion
/blog/microprofile/2021/03/31/microprofile-config-2.0.html=/blog/2021/03/31/microprofile-config-2.0.html
```
This `,` looks like a bug/typo from a past commit

kinueng (approved):
LGTM

blog-redirects.properties
mbroz2: it is not; that is partially what the redirect is there for

Issue #1566, opened 2021-07-19
Open Liberty with OpenJ9 blog


gljrobinson:

gljrobinson:
@yeekangc is this related to any of the blogs on the community site?

gljrobinson:
Perhaps a walk through of using Liberty with Open J9 through IBM Semeru Runtimes, point to other existing resources around advantages.

yeekangc:
@jamiecoleman92 indicated he will get started on this blog post this week.

jamiecoleman92:
[Open Liberty & Semeru Blog.docx](https://github.com/OpenLiberty/blogs/files/8842903/Open.Liberty.Semeru.Blog.docx)
Here is the latest version of the blog post. @yeekangc if you could give it a quick review that then I will convert it to ASCII Doc and make a PR.

yeekangc:
@jamiecoleman92, you can submit a PR with the draft and we can iterate on it.

Can we dive deeper into the technical aspects and advantages of OpenJ9 as a VM?  e.g., how it goes well with Liberty for cloud-native workloads?  Is there any perf data we can share?  Cc @vijaysun-omr.

Also, I am not sure we need to repeat what is Open Liberty since this is the Open Liberty blog.

It's "IBM Semeru Runtimes" and we should use it as such (and correctly).  Cc @rhagarty in case he has comments what more we can consider in terms from the perspective of OpenJ9/IBM Semeru Runtimes.

@lauracowen, can you take a look at the draft and offer your thoughts too when you have a moment?

vijaysun-omr:
I agree with YK, the JDK distribution is what is referred to as "IBM Semeru Runtimes", the JVM's name is Eclipse OpenJ9. We should be clear in expressing that so that users also see the right terminology (in Semeru's case, it is still fairly new).

There are some performance results with the OpenJ9 JVM shown on openliberty.io itself and we should probably link to that homepage. There are also other blogs such as https://openliberty.io/blog/2019/10/30/faster-startup-open-liberty.html that talk about the one second startup time that you mention in the blog post.

vijaysun-omr:
There is also https://www.eclipse.org/openj9/performance/ that lists some information about Eclipse OpenJ9's characteristics including some quotes from people who have tried it.

jamiecoleman92:
Thanks for the feedback @yeekangc & @vijaysun-omr 
I will incorporate the above and create a PR with everything in markdown.
[...]

PR #2504, opened 2022-08-19
Staging to prod- OL/Semeru post

dmuelle:
#1566 

looks good on [staging](https://staging-openlibertyio.mybluemix.net/blog/2022/08/19/open-liberty-semeru-performance.html):
Screen Shot 2022-08-19 at 12 14 25 PM


mbroz2 (approved):
`open-graph-image: ` should be update to one of the 4 images, I'd go with the startup time bar graph

PR #2540, opened 2022-09-07
Update with 2022 performance collateral

tpickett1:
Updating the performance charts and tables.  This update is for the original blog post by Graham Charters -Microservice performance that saves you money


dmuelle (approved):
LGTM

PR #2554, opened 2022-09-14
Mp perf collateral update

tpickett1:
Update this blog post with newest Performance data/charts.


tpickett1:
If there is a why we could tweet or other form of external notice that this post has updated performance data, it would be great.

mbroz2 (changes requested):
This new performance data should be moved to a new blog post.  The old 2020-10-21-memory-footprint-throughput.adoc post can be updated to point people at the new post via a link.

tpickett1:
ok, I'll work on the new post with Graham, and once that is done I'll update the old blog post with a link to the new one.
Thanks.

gljrobinson:
Hi @tpickett1, as you published a separate blog with Graham, is this PR ok to be closed?

tpickett1:
yes, it can be closed

PR #2565, opened 2022-09-19
Update to use getting-started guide

tjwatson:
This guide is already setup to containerize the application so it makes the setup easier to demonstrate InstantOn.


dazavala (approved):

tjwatson:
For #2585

Issue #2585, opened 2022-09-26
Open Liberty release InstantOn blog post for 220011-beta


tjwatson:
This will be a stand alone blog post about Open Liberty InstantOn.

gljrobinson:
This is now published.

PR #2590, opened 2022-09-27
Normalize performance testing results

dettmoney:
The chart and description need to be updated to normalize results, with the baseline being 100


tjwatson:
For #2585

PR #2695, opened 2022-10-28
Merge tag updates back to staging

lauracowen:
oops, sorry, I committed to prod rather than staging first.

### DIFF
--- a/blog_tags.json
+++ b/blog_tags.json
@@ -2,7 +2,7 @@
     "blog_tags": [
         {
             "name": "announcements",
-            "posts": ["22.0.0.11", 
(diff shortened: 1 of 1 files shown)


mbroz2:
I skipped the tags for 22.0.0.12-beta initially because there was a [bug introduced into the website code](https://github.com/OpenLiberty/openliberty.io/issues/2891) that wasn't rendering tags for posts with periods.  I did a fix for it yesterday and Kin delivered it so it's working now

mbroz2 (approved):

lauracowen:
Yes, I was tracking that bug because I spotted the problem yesterday, which is how I noticed this one was missing its tags today. I figured it was an easy fix to add the tags this morning. :)

Issue #2675, opened 2022-10-26
Create blog about using the Paketo Buildpack for Liberty in IBM Cloud Code Engine


kevin-ortega:
There is interest in using the Paketo buildpack in Code Engine (CE).  There is no documentation on how to use the buildpack in CE and the CE team does not currently have the bandwidth to create the doc.   

We will create a blog to document how to use the buildpack in CE while we wait for the CE documentation to be updated.

dmuelle:
published at https://openliberty.io/blog/2022/11/14/using-codeengine-with-paketo-buildpack-for-liberty.html

closing as completed, thanks

PR #2750, opened 2022-11-14
publish Paketo post

dmuelle:
#2675 

Looks good on [staging](https://blogs-staging-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/blog/2022/11/14/using-codeengine-with-paketo-buildpack-for-liberty.html)
Screen Shot 2022-11-14 at 3 33 42 PM


mbroz2 (approved):
LGT (we probably could have linked "Supports any valid Liberty feature." to https://openliberty.io/docs/latest/reference/feature/feature-overview.html but that's minor)

PR #2814, opened 2022-12-16
Edits to 2022 year in review post

gljrobinson:
Two sentences edited and an image added to the 2022 year in review post


rhagarty (approved):
Good catches. Thanks Grace

PR #2921, opened 2023-02-15
Testing action for projects

jakub-pomykala:
Changed variable name for GitHub token in projects action.


mbroz2 (approved):
Interesting, I wasn't aware that the new GH projects can't be scoped to the repo, and are instead at the org level.

PR #3024, opened 2023-03-27
Slack pagination

mbroz2:
break apart slack msgs to not exceed the 4000 char limit.


kinueng (approved):
LGTM, though my sight reading of python is a bit weak nowadays. It looks that the "data" is around a GitHub issue's text that is already public so there should not be any concerns.

PR #3054, opened 2023-04-12
Update mpTelemetry blog

Yas-Franco:
The blog no longer needs to mention the beta release as the feature is now in GA


dmuelle:
Need to target draft, then staging before prod to keep branches in sync

PR #3206, opened 2023-07-13
Instanton chinese

Emily-Jiang:
InstantOn Chinese translation


lauracowen (approved):
Looks ok. I can't read Chinese but it doesn't look like there's anything obviously wrong with the structure/tagging/formatting. You can check that for sure when it's built anyway.

yeekangc:
@tjwatson recently updated the beta blog posts to reflect what's in the GA.  I believe we have mentioned that the capability has GAed in this Chinese translation and is pointing to the latest GA blog post, which is nice.

Then, I believe there are still differences with what Tom updated: https://openliberty.io/blog/2022/09/29/instant-on-beta.html.  Will be nice to make the Chinese translation consistent with the latest version of this beta blog post?

Just thought to note.

Emily-Jiang:
ah, I see. I will do the similar updates. Thanks @yeekangc

PR #3215, opened 2023-07-18
Fix ja path

natalie-bernhard:
Fix the ja path in the front matter of the English blog post for 23.0.0.5 release blog


dmuelle (approved):
LGTM

PR #3389, opened 2023-09-26
Liberty olo with rh sso

rumanaHaque:
PR to check in blog about securing Open Liberty App using RH-SSO to draft.


lauracowen:
@rumanaHaque Could you just merge this to draft if you're ready for a review, pls (I think you should be able to do that without reviewers, or just get someone in your team to approve that the file seems ok and no obvious technical problems that would cause the build to fail? Then it will automatically build to the [draft site](https://draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/blog/) so you and reviewers can review it there. Create another PR to `staging` branch for review feedback.

PR #3437, opened 2023-11-06
Update blog tags

SteveSamJacob19:


blog_tags.json
```diff
@@ -333,7 +333,7 @@
                       "alerts-slack-prometheus-alertmanager-open-liberty"]
         },
         {
-            "name": "Jakarta EE",
+            "name": "jakartaee",
```
steven1046: can you hyphenate this one so the tags with spaces before are consistent?

steven1046 (changes requested):

steven1046 (approved):

PR #3563, opened 2024-02-08
Update blog_tags.json

lauracowen:
Created new `user story` tag to group posts about Open Liberty uses.

This look ok to you, @mbroz2 ?

### DIFF
--- a/blog_tags.json
+++ b/blog_tags.json
@@ -520,7 +520,7 @@
                       "jakartaee9-data-source-transactions-20008-beta", "testing-database-connections-REST-APIs"]
         },
         {
-            "name": "user story",
+            "name": "case-study",
             "posts": [ "Primeur", "production-experience-open-liberty"]
         },
         {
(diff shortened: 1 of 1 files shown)


blog_tags.json
```diff
@@ -519,6 +519,10 @@
                       "database-ssl-primer", "pre-populating-db-connections",
                       "jakartaee9-data-source-transactions-20008-beta", "testing-database-connections-REST-APIs"]
         },
+        {
+            "name": "user story",
```
mbroz2: Our naming convention for blog tags is to use dashes (`-`) and not spaces (and all lower case letters, though not an issue here).  
```suggestion
            "name": "user-story",
```

Also, when I read 'user story' the first thing that comes to mind is Agile process user story.  Given the prevalence of this term in the industry, I'm worried some might get the wrong idea behind the label.  Perhaps another term?
testimonial, reference, user-reference, use-case, case-study, end-user, liberty-user,

mbroz2 (changes requested):

blog_tags.json
lauracowen: This is why I wanted your review. :) Thanks. I've changed to `case-study`.

mbroz2 (approved):

PR #3723, opened 2024-04-25
Argo tags

dmuelle:


danielguinan1 (commented):
Is testcontainers-guide tag for something else?

dmuelle:
The testcontainers tag is just showing up because it's getting moved down a line with the addition of the argo post tags, it was already in this section of the file

mtamboli (approved):
This should be fine.

PR #4056, opened 2024-10-01
simplify and differentiate examples

covener:
The preferred algorithm value was the syntax, not a valid value. It's also relatively obscure function.

Simplify the examples and make the content-type one dedicated in a third example.


dmuelle:
Hi @covener - this needs to go to draft--->staging before prod. I'll merge here so the cvhange can be reviewed on the draft site before moving to staging and prod. I opened #4058 to track progress.

PR #4088, opened 2024-10-09
simplify and differentiate examples

dmuelle:
The preferred algorithm value was the syntax, not a valid value. It's also relatively obscure function.

Simplify the examples and make the content-type one dedicated in a third example.


dmuelle:
updated post on draft: https://blogs-draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/blog/2020/04/22/http-response-compression.html

covener (approved):
LGTM including the rendered draft.

PR #4149, opened 2024-11-19
Contribution from Takakiyo-san and update to the original version wit…

una-tapa:
…h available language

This has been translated and reviewed by the Japan team.  
Requesting to publish.


dmuelle (approved):
LGTM

PR #4198, opened 2024-12-09
Publish Cloudant post and Japanese 24.0.0.12

### LINKED ISSUE #4152: Access Cloudant Client with Open Liberty using CDI
The oroginal Cloudant client which is the `cloudant-1.0` feature in Open Liberty was built around stopped development ~2 years ago. But we still want cloudant to be usable with Liberty, so in order to stabilize that, as an alternative, we created a sample application that just demonstrates to users how to pull in the cloudant client and use it directly in their application. 

This blog will be in the style of a guide which users can refer for using cloudant feature in their application.

dmuelle:
Fixes #4152


revijay (approved):
Cloudant file changes look good

PR #4213, opened 2024-12-16
Japanese blog two year celebration

una-tapa:
Hi, @dmuelle 

This article marks the second anniversary of the Japanese Open Liberty blog. Unfortunately, we encountered an issue where the link https://github.com/una-tapa could not be removed after being merged into the draft. We would appreciate assistance in removing this, as well as the word "and" which we find to be extraneous.

image

Apart from the above, the draft has been reviewed and approved by everyone involved. We are hoping it to be published by December 20th to coincide with our participation in the WebSphere Advent Calendar event over there. 

Thank you in advance!


dmuelle (approved):
LGTM

PR #4227, opened 2025-01-21
Update SB3 WAR post to highlight use of Spring Boot application props

scottkurz:
Continuing https://github.com/OpenLiberty/blogs/pull/4221

PR #4221, opened 2025-01-17
Update SB3 WAR post to highlight use of Spring Boot application props

scottkurz:
Responding to this Q&A for liberty-maven-plugin, https://github.com/OpenLiberty/ci.maven/discussions/1859, I realized this article could be improved by specifically mentioning how the Spring Boot "optimized" deployment lets you use Spring Boot-centric config like application.properties.

I made a small rework to the post to show it.

WDYT about updating it with this PR?    

(Note I haven't yet merged in the new code to the linked repo.. I'd do that when done). 

Thanks

PR #4235, opened 2025-01-23
Sb3 0125 update

scottkurz:
follow-up of #4221 and #4227


scottkurz:
I think it's OK if I merge this myself to kick off the draft build.

PR #4481, opened 2025-08-11
Update prod

navaneethsnair1:
Link to staging site [here](https://blogs-staging-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/blog/)

PR #4483, opened 2025-08-13
Update prod

navaneethsnair1:
Addressing changes suggested by @ayoho on 25.0.0.8 GA blog ([here](https://github.com/OpenLiberty/blogs/pull/4481#discussion_r2270474998))

### DIFF
--- a/posts/2025-08-12-25.0.0.8.adoc
+++ b/posts/2025-08-12-25.0.0.8.adoc
@@ -227,13 +227,13 @@ For more information about MicroProfile OpenAPI, see:
 |3.7
 |Bypass security
 |17.0.0.3 - 25.0.0.7
-|Affects the `servlet-3.1`, `servlet-4.0`, `servlet-5.0`, and `servlet-6.0` features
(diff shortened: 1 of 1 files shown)


ayoho (commented):
Looking at this again, I'm not sure how this is supposed to render. If the `feature:servlet-3.1[]` syntax is supposed to link to the `servlet-3.1` feature documentation, then something else needs to be fixed to make sure that works. If these are **not** expected to be links, then the `feature:` prefix should also be taken out.

navaneethsnair1:
> Looking at this again, I'm not sure how this is supposed to render. If the `feature:servlet-3.1[]` syntax is supposed to link to the `servlet-3.1` feature documentation, then something else needs to be fixed to make sure that works. If these are **not** expected to be links, then the `feature:` prefix should also be taken out.

Dont know why was it given like that in CVE list in github [here](https://github.com/OpenLiberty/docs/blob/draft/modules/ROOT/pages/security-vulnerabilities.adoc). I will update them referring previous examples.

PR #4705, opened 2026-02-16
26.0.0.2-beta blog update

navaneethsnair1:
https://blogs-draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/blog/2026/02/10/26.0.0.2-beta.html

### DIFF
--- a/posts/2026-02-10-26.0.0.2-beta.adoc
+++ b/posts/2026-02-10-26.0.0.2-beta.adoc
@@ -33,7 +33,7 @@ See also link:{url-prefix}/blog/?search=beta&key=tag[previous Open Liberty beta 
 // // // // // // // // 
 [#mcp]
 == Model Context Protocol Server 1.0 updates
-The link:https://modelcontextprotocol.io/docs/getting-started/intro[Model Context Protocol (MCP)] is an open standard that enables AI applications to access real-time information from external sources. The Liberty MCP Server feature `mcpServer-1.0` allows developers to expose the business logic of their applications, allowing it to be integrated into agentic AI workflows.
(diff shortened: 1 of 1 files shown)


posts/2026-02-10-26.0.0.2-beta.adoc
```diff
@@ -204,50 +208,22 @@ The following additional configuration is required in the `server.xml` (also see
...
-        logger.info("Processing request [" + requestId.asString() + "] with input: ");
-        ...
-    }
-}
-----
-
-=== MCP: Tools can access metadata that is sent by the client in the `_meta` field
-
-The link:https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta[MCP specification] specifies that the `_meta` property/parameter is reserved by MCP to allow clients and servers to attach additional metadata to their interactions.
+It is also possible to add multiple roles to tools: `@RolesAllowed("Admins, Moderators")`. Doing so grants access to users who have either role.
```
Azquelt:
```
It is also possible to add multiple roles to tools: `@RolesAllowed("Admins, Moderators")`. Doing so grants access to users who have either role.

If a role is used in a `@RolesAllowed` annotation, but isn't mentioned in the `server.xml` file, that just means that no user has that role. If it's the only role that's allowed to call the tool, then no users will be able to call the tool.
```

Can we swap these two statements, and change the one about a role not defined in the server.xml.

Things I want to avoid saying when talking about a role not defined in server.xml:
- "the server.xml file excludes a role"
  - "excludes a role" sounds like there's config in server.xml which excludes a role, which isn't the case
- "the role or its permissions"
  - we don't really have a concept of "permissions"
  - we declare that certain roles are allowed to access certain resources
  - upon attempting to access a resource, we check whether the user has any of the roles that are allowed to access it
- "the user will be denied access" (without clarification)
  - if multiple roles are allowed to access the tool, users with another role can still access it
  - this is why I would like to put the statement about multiple roles first

PR #4739, opened 2026-03-06
26.0.0.3-beta

IsmathBadsha:
[Issue](https://github.com/OpenLiberty/open-liberty/issues/34061)


posts/2026-03-10-26.0.0.3-beta.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+When a client calls a tool, the object that is returned by the tool method is converted to a `ToolResponse`, which usually contains one or more `Content` objects. The `ToolResponse` maps directly to the result returned in the response.
+
+If you need more control over the response from your tool, you can return a `ToolResponse` or `Content` object directly. Now it is also possible to register encoders to control how other objects are converted into a response.
+
+* Use `ToolResponseEncoder` to convert an object into a `ToolResponse`, which gives you complete control over the whole response.
+* Use `ContentEncoder` only when you need to convert an object into a `Content` that is included in the response. You can also return a list of objects, and each object is individually converted into a `Content` and included in the response.
```
Azquelt, line 44: I prefer "Use ContentEncoder when you only need to convert an object into a `Content`" (the other changes are fine).

To add some context: A tool response contains one or more content object. You can use `ToolResponseEncoder` to create a whole tool response (including one or more content objects), or you can use `ContentEncoder` to create a content object, and then the server will put that content object in a tool response for you.

In the original, I use "only" to indicate that `ContentEncoder` does a more limited thing than `ToolResponseEncoder`. In the new version it sounds like we're placing a limitation on when `ContentEncoder` should be used.

Removing the word "only" here would also be fine.

Azquelt (dismissed):
I have ~one~ two comments, the rest looks good.

posts/2026-03-10-26.0.0.3-beta.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+// // // // DO NOT MODIFY THIS COMMENT BLOCK <GHA-BLOG-TOPIC> // // // // 
+// Blog issue: https://github.com/OpenLiberty/open-liberty/issues/XXXXX
+// Contact/Reviewer: martindrozdz
+// // // // // // // // 
+[#encoders]
+== Encode responses with ContentEncoder and ToolResponseEncoder
+
+The link:https://modelcontextprotocol.io/docs/getting-started/intro[Model Context Protocol (MCP)] is an open standard that enables AI applications to access real-time information from external sources. The Liberty MCP Server feature (`mcpServer-1.0`) allows developers to expose the business logic of their applications, allowing it to be integrated into agentic AI workflows.
+
```
Azquelt, line 38: This paragraph is a little out of place here, since it relates to MCP as a whole, rather than the `Encode responses with ContentEncoder and ToolResponseEncoder` section.

I'd consider putting it above the heading maybe?

PR #4947, opened 2026-06-15
Update prod

navaneethsnair1:


blog_tags.json
```diff
@@ -2,7 +2,7 @@
     "blog_tags": [
         {
             "name": "announcements",
-            "posts": ["26.0.0.5", "26.0.0.5-beta", 
+            "posts": ["26.0.0.6-beta", "26.0.0.5", "26.0.0.5-beta",
```
ayoho: This should reference 26.0.0.6 now. There should be a reference to 26.0.0.6-beta as well - we might have missed that when that blog was released. But this array should include both 26.0.0.6 and 26.0.0.6-beta now.

Similar comment for all the tags - make sure 26.0.0.6-beta and 26.0.0.6 are in the appropriate places.

posts/2026-06-16-26.0.0.6.adoc
```diff
@@ -0,0 +1,308 @@
...
+
+* link:https://github.com/OpenLiberty/open-liberty/issues/34863[featureUtility exits with return code of 0 with malformed server.xml]
+* link:https://github.com/OpenLiberty/open-liberty/issues/34857[FFDC for AlreadyBound exception in EJBRemoteRuntimeImpl during application start]
+* link:https://github.com/OpenLiberty/open-liberty/issues/34802[fix AES encryption issues with featureUtility]
+* link:https://github.com/OpenLiberty/open-liberty/issues/34773[java9.options not loaded on Windows when JAVA_HOME not set]
++
+On Windows, the server.bat script fails to load java9.options (and java24.options) when JAVA_HOME is not explicitly set, even though Java can be successfully located via JRE_HOME or WLP_DEFAULT_JAVA_HOME environment variables. The script uses these variables to find and execute the java binary, but never sets JAVA_HOME from them. Without JAVA_HOME set, the Java version detection logic (lines 768-795) cannot locate the release file or lib\modules directory, preventing java9.options from being loaded.
```
ayoho: Was it intentional to include the content from the "Describe the bug" section in the corresponding issues here? And why is it done for the bugs from here below, but not for the initial few bugs in this list?

navaneethsnair1: Other descriptions were either absent or looked vague for me. I can delete those lines for now and incorporate them later.

blog_tags.json
navaneethsnair1: I was not sure if release with primarly bug fixes could be added within announcements  tag, etc. I will add that right away

PR #5026, opened 2026-08-05
MCP Standalone Blog Updated - Fix the blog release date

TheoGkoumas:
This blog post will be correct for the 26.0.0.9-beta, so the eGA date is 11-Aug, and the beta blog will probably go out on 25-Aug. Thus, a change on the post date to the 25th of August is reasonable.


Azquelt:
[From the readme](https://github.com/OpenLiberty/blogs#editors-editing-and-publishing-a-post), it looks like we shouldn't add tags here, that will be done by one of the editors when they merge it to staging.

Azquelt:
The file name will become the URL, so I think it should be something that better reflects what users will get from it. I don't think it should have `standalone`, `blog` or `updated` in there.

Looking back through the posts directory, there's a bit of variation. I'm thinking maybe `2026-08-25-mcp-tools.asciidoc`, though I'm open to other suggestions.

Azquelt (approved):
I would squash these commits, but otherwise it's ready to merge

TheoGkoumas:
Commits were squashed to [3dce894](https://github.com/OpenLiberty/blogs/commit/3dce8949cbfa570437f1a6cf0d70728cb5ffe7e7)

PR #5043, opened 2026-08-27
Fix/2026 08 25 mcp tools

devalibm:
- **Feature name**: Replaced all occurrences of the incorrect feature name `mcpServer-1.0` with the correct `mcp-1.0`.
- **Maven dependency**: Replaced with a clean `provided`-scoped Maven Central dependency for the Liberty MCP API

### DIFF
--- a/posts/2026-08-25-mcp-tools.adoc
+++ b/posts/2026-08-25-mcp-tools.adoc
@@ -6,7 +6,7 @@ author_picture: https://avatars3.githubusercontent.com/TheoGkoumas
 author_github: https://github.com/TheoGkoumas
 seo-title: Support for the Model Context Protocol within Open Liberty - OpenLiberty.io
 seo-description: Learn how to use the MCP Server feature within your Liberty application to allow your business logic to be utilized within an agentic AI workflow.
-blog_description: Learn how to use the MCP Server feature within your Liberty application to allow your business logic to be utilized within an agentic AI workflow.
+blog_description: "Learn how to use the MCP Server feature within your Liberty application to allow your business logic to be utilized within an agentic AI workflow."
 open-graph-image: https://openliberty.io/img/twitter_card.jpg
 open-graph-image-alt: Open Liberty Logo
 ---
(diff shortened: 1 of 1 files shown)


Azquelt:
@devalibm The PR diff now looks right and has no conflicts, but we can't build your update on the `draft` branch. That would result in everything from `draft` going live when your change is merged into `prod`.

Azquelt:
It doesn't look like this was finished, it was supposed to go out with the GA blog because the GA blog links to it.

posts/2026-08-25-mcp-tools.adoc
```diff
@@ -6,7 +6,7 @@ author_picture: https://avatars3.githubusercontent.com/TheoGkoumas
 author_github: https://github.com/TheoGkoumas
 seo-title: Support for the Model Context Protocol within Open Liberty - OpenLiberty.io
 seo-description: Learn how to use the MCP Server feature within your Liberty application to allow your business logic to be utilized within an agentic AI workflow.
-blog_description: Learn how to use the MCP Server feature within your Liberty application to allow your business logic to be utilized within an agentic AI workflow.
+blog_description: "Learn how to use the MCP Server feature within your Liberty application to allow your business logic to be utilized within an agentic AI workflow."
```
Azquelt, line 9: Why did we put quotes around this?

posts/2026-08-25-mcp-tools.adoc
Azquelt, line 458: The change here makes sense, but the example isn't very good, we should raise a new issue to rewrite it to use restfulWS client or MP Rest Client, both of which have an async call API.

posts/2026-08-25-mcp-tools.adoc
Azquelt: It looks like some blog posts use quotes and some don't but I can't see an obvious pattern. We can leave as-is and check it renders correctly on the draft site.

Azquelt (approved):

Azquelt:
Forced-push to sign commits properly. No changes.

posts/2026-08-25-mcp-tools.adoc
devalibm: Sure, I will remove it

Issue #5068, opened 2026-09-18
Fix the config examples in the MCP tools blog post


Azquelt:
The examples in https://openliberty.io/blog/2026/08/25/mcp-tools.html still use `<mcpServer>` (which we used in the beta) instead of `<mcp>`.

PR #5069, opened 2026-09-18
Fix MCP config examples

Azquelt:
Config examples should use `<mcp>`.

Reference: https://openliberty.io/docs/latest/reference/config/application.html#mcp

For #5068


IssacAX123 (approved):
LGTM

